PR #3, opened 2022-05-18
Bring over our composite actions from Khan/perseus

jaredly:
## Summary:
They'd been copied over from graphql-flow, but I figured the Khan/perseus ones might have more recent changes.
The next PR will contain our build & publish scripts.

I also created a 'check-for-changeset' action that I noticed in the perseus pr-actions workflow :)

Issue: https://khanacademy.atlassian.net/browse/FEI-4550

## Test plan:
🤞 not usable/testable quite yet!


actions/filter-files/action.yml
```diff
@@ -0,0 +1,37 @@
...
+    description: 'The jsonified list of files that match'
+    value: ${{ steps.result.outputs.result }}
+runs:
+  using: "composite"
+  steps:
+    - uses: actions/github-script@v6
+      id: result
+      with:
+        script: |
+          const extensionsRaw = "${{ inputs.extensions }}";
```
jeresig, line 24: I wonder if we should have a policy (in a future change) of breaking out inline scripts to be imported from another file? That way we can lint and test them, for example: https://github.com/actions/github-script#run-a-separate-file

jeresig (approved):
This looks awesome!

actions/filter-files/action.yml
jaredly: on interesting, I didn't know that was an option! Sounds like a good policy.

jeresig: Always been possible, I believe!

actions/full-or-limited/action.yml
```diff
@@ -0,0 +1,28 @@
+name: Full or Limited
```
kevinbarabash, line 1: Thanks for abstracting this into an action.

PR #5, opened 2022-05-19
Add eslint & prettier

jaredly:
## Summary:
It's very simple for the moment because we're not doing any transpilation. We might decide to adopt typescript or something if we start to have more complicated actions.

Issue: XXX-XXXX

## Test plan:
`yarn lint` should work!
Also see the github action doing its thing.


.eslintrc.js
```diff
@@ -0,0 +1,28 @@
+/**
+ * This is a very stripped-down eslintrc, because @khanacademy/eslint-config
+ * assumes you're using react & flow (and requires a lot of dependencies along
+ * with it).
+ */
+const khan = require("@khanacademy/eslint-config");
+
+// Only keep the rules that have no prefix (so are included in eslint core)
+// and the prettier rule.
```
kevinbarabash, line 9: I had a look at `khan/eslint-config-khan` and see that everything's listed as a peer dep.  I wonder if we should change that.

package.json
```diff
@@ -4,7 +4,16 @@
     "workspaces": [
         "actions/*"
     ],
+    "scripts": {
+        "lint": "eslint --ext .js,.mjs .",
```
kevinbarabash, line 8: I didn't realize you pass a comma separated list to `--ext`, cool!

kevinbarabash (approved):
LGTM. :shipit:

PR #6, opened 2022-05-19
Add major version tags, to allow implicit upgrades for patch & minor versions

jaredly:
## Summary:
it'd be a hassle to have to change all our workflows for patch versions of these.
Now you can do `uses: Khan/actions@filter-files-v1` and it will get all 1.x.x versions.

Issue: XXX-XXXX

## Test plan:
```bash
➜ g tag -d full-or-limited-v0.0.0
Deleted tag 'full-or-limited-v0.0.0' (was 4e86abe)

➜ g push origin :full-or-limited-v0.0.0
To github.com:Khan/actions.git
 - [deleted]         full-or-limited-v0.0.0

➜ node utils/run-publish.mjs
From .
 * branch            lints      -> FETCH_HEAD
To github.com:Khan/actions.git
 * [new tag]         full-or-limited-v0.0.0 -> full-or-limited-v0.0.0
remote: warning: Deleting a non-existent ref.
To github.com:Khan/actions.git
 - [deleted]         full-or-limited-v0
To github.com:Khan/actions.git
 * [new tag]         full-or-limited-v0 -> full-or-limited-v0
```


utils/publish.mjs
```diff
@@ -18,16 +18,20 @@ export const checkTag = (tag) => {
...
         `git commit -m publish`,
         `git remote add origin ${origin}`,
         `git tag ${tag}`,
+        `git tag ${majorTag}`,
     ];
     if (!dryRun) {
         cmds.push(`git push origin ${tag}`);
+        // This will succeed with a warning if the major tag doesn't exist
+        cmds.push(`git push origin :refs/tags/${majorTag}`);
+        cmds.push(`git push origin ${majorTag}`);
```
kevinbarabash, line 34: So this overwrite the major tag with whatever the most updated to minor.patch version within that major version is?  Nice solution.

jeresig (approved):
Very nice, thank you! @kevinbarabash I was talking with Jared about this and I really think that either folks want to pin to a specific version or they want major updates, but we don't need to care about other cases, I think. If we ever do we can add those release tags, too.

PR #7, opened 2022-05-19
Add a shared action for gerald, and use it!

jaredly:
## Summary:
Testing this out

Issue: XXX-XXXX

## Test plan:


.github/workflows/gerald-pr.yml
```diff
@@ -0,0 +1,39 @@
...
+jobs:
+  gerald:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: ./actions/gerald
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          admin-token: ${{ secrets.KHAN_ACTIONS_BOT_TOKEN }}
+      # - name: Get All Changed Files
```
jeresig, line 15: Whoops, can delete this!

actions/gerald/action.yml
```diff
@@ -0,0 +1,34 @@
+name: Gerald
```
jeresig, line 1: Gerald has a few different actions - I think we may want to call this "gerald-pr" - there would also be a "gerald-comments" and "gerald-push".

jeresig (approved):

PR #8, opened 2022-05-19
Use the local versions of our actions in our workflows, and add jest

jaredly:
## Summary:
By referencing the actions locally, we can test out changes in the same pull-request, instead of waiting until landing (when the new version would get published).
I also noticed a bug in build.mjs (using `#` instead of `@`, which will result in an error in github actions), so I added a test for processActionYml.

Issue: XXX-XXXX

## Test plan:
`yarn test`


utils/build.mjs
```diff
@@ -31,8 +31,8 @@ export const processActionYml = (
     const replacements = [];
     Object.keys(packageJsons[name].dependencies).forEach((depName) => {
         replacements.push({
-            from: new RegExp(`\\buses: ${depName}\\b`, "g"),
-            to: `uses: ${monorepoName}#${depName}-v${packageJsons[depName].version}`,
+            from: new RegExp(`\\buses: ./actions/${depName}\\b`, "g"),
```
jeresig, line 34: I think you need to escape the .
```suggestion
            from: new RegExp(`\\buses: \\./actions/${depName}\\b`, "g"),
```

jeresig (approved):

PR #10, opened 2022-05-24
Cache ~/.cache/Cypress if it's there

jaredly:
## Summary:
it'll no-op if we're not using cypress in the project, so no harm done.

Issue: XXX-XXXX

## Test plan:
🤞


kevinbarabash (approved):
LGTM.  Thanks for adding cypress caching. 🎉

PR #11, opened 2022-05-25
Bring "@jaredly/get-changed-files" into our actions monorepo, stripping it down to only what we need

jaredly:
## Summary:
We've unified around json encoding to deal with spaces, and we only ever ask for added_modified and renamed, so I merged them together.

This is my first time trying `actions/github-script` with a local require, and it turns out it's relative to the repo base (which makes sense). Because of this, I had to change `build.mjs` to turn `require('./actions/get-changed-files/index.js')` into `require('./index.js')` when publishing these actions individually.

Issue: XXX-XXXX

## Test plan:
See the check on this one!

image

### DIFF
--- a/actions/get-changed-files/index.js
+++ b/actions/get-changed-files/index.js
@@ -1,0 +1,73 @@
+// Get the changed files for a pull-request or push
+
+const getBaseAndHead = (context, core) => {
+    switch (context.eventName) {
+        case "pull_request_target":
+        case "pull_request":
+            return [
+                context.payload.pull_request?.base?.ref,
+                context.payload.pull_request?.head?.sha,
+            ];
+        case "push":
+            return [context.payload.before, context.payload.after];
+        default:
+            core.setFailed(
+                `This action only supports pull requests and pushes, ${context.eventName} events are not supported. ` +
+                    "Please submit an issue on this action's GitHub repo if you believe this in correct.",
+            );
+    }
+};
+
+module.exports = async ({github, context, core}) => {
+    const absolute = !!core.getInput("absolute", {required: false});
+
+    const [base, head] = getBaseAndHead(context, core);
+    core.info(`Base: ${base}\nHead: ${head}`);
+
+    // Ensure that the base and head properties are set on the payload.
+    if (!base || !head) {
+        core.setFailed(
+            `The base and head commits are missing from the payload for this ${context.eventName} event.`,
+        );
+        return;
+    }
+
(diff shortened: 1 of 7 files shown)


actions/get-changed-files/index.js
```diff
@@ -0,0 +1,73 @@
...
+
+    // Ensure that the request was successful.
+    if (response.status !== 200) {
+        core.setFailed(
+            `The GitHub API for comparing the base and head commits for this ${context.eventName} event returned ${response.status}, expected 200.`,
+        );
+    }
+
+    const files = response.data.files.filter((file) =>
+        ["added", "modified", "renamed"].includes(file.status),
```
somewhatabstract, line 52: Nice! Thanks for combining the renames with the others.

utils/build.mjs
```diff
@@ -28,8 +28,8 @@ export const processActionYml = (
     actionYml,
     monorepoName,
 ) => {
-    const replacements = [];
-    Object.keys(packageJsons[name].dependencies).forEach((depName) => {
+    const replacements = [{from: new RegExp(`\\./actions/${name}/`), to: "./"}];
```
somewhatabstract, line 31: Can you add a comment here explaining what this is for? Probably can just copy what you wrote in the PR description.

utils/build.mjs
```diff
@@ -28,8 +28,8 @@ export const processActionYml = (
...
+    Object.keys(packageJsons[name].dependencies || {}).forEach((depName) => {
```
somewhatabstract, line 32: NIT: Use `??` instead of `||`

somewhatabstract (approved):
LGTM

Would be useful to include some additional commentary in the source to explain some bits, but overall, nothing blocking.

utils/build.mjs
```diff
@@ -28,8 +28,13 @@ export const processActionYml = (
     actionYml,
     monorepoName,
 ) => {
-    const replacements = [];
-    Object.keys(packageJsons[name].dependencies).forEach((depName) => {
+    // This first replacement is to rewrite local requires, in the case where we have
+    // a github-script action with e.g. `require('./actions/my-action/index.js')`. Writing
+    // it this way means it will work in this repo without publishing (so our workflows can
+    // use it directly), and then we do this replacement when publishing so it will work there
+    // too.
```
somewhatabstract, line 35: Thank you!

somewhatabstract (approved):

PR #12, opened 2022-05-31
Add `-f` to `git fetch --tags` in the publish script to allow clobbering

jaredly:
## Summary:
Because we rewrite our major version tags (filter-files-v1 for example) on every patch & minor version publish, tags will move around, and `-f` is needed if you have different tags locally.

Issue: XXX-XXXX

## Test plan:


utils/publish.mjs
```diff
@@ -56,7 +56,7 @@ export const collectPackageJsons = (packageNames) => {
 };
 
 export const publishAsNeeded = (packageNames, dryRun = false) => {
-    execSync(`git fetch --tags`);
+    execSync(`git fetch --tags -f`);
```
jeresig, line 59: Probably worth adding a comment here essentially copying in the PR message so that future folks know why this is important!

jeresig (approved):

somewhatabstract (approved):
This looks fine to me. Thanks for adding the explanation of why this is needed.

PR #14, opened 2022-05-31
Fix resolution of local .js files in the built action

jaredly:
## Summary:
So I tried using get-changed-files, and it broke! apparently rewriting it to `require("./index.js")` wasn't the right way to go about it.
Then I discovered `${{ github.action_path }}`, which is the right way to do this.

Issue: XXX-XXXX

## Test plan:
`yarn jest`
Also, see it working here https://github.com/Khan/actions/runs/6678289285?check_suite_focus=true, I've set gerald-pr to use Khan/actions@get-changed-files-v0.0.0-test, which I published with this change.

### DIFF
--- a/.changeset/thick-lobsters-obey.md
+++ b/.changeset/thick-lobsters-obey.md
@@ -1,0 +1,5 @@
+---
+"get-changed-files": patch
+---
+
+Fix resolution of the index.js file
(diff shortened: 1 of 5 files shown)


jeresig (approved):
Ohh - very interesting! That's helpful!

PR #18, opened 2022-06-01
Add 'invert' option to filter-files

jaredly:
## Summary:
This allows you to filter-out paths instead of filtering for them. To used in check-for-changeset.

Issue: XXX-XXXX

## Test plan:
see example workflow


jeresig (approved):
Nice!

PR #21, opened 2022-06-01
Add a 'directories' option to get-changed-files, for prefiltering by directory

jaredly:
## Summary:
I thought about allowing regexes, or minimatch or something like that, but I don't think we need to get that complicated. If you want to really drill down, that's what filter-files is for.

Issue: XXX-XXXX

## Test plan:
See the node-ci workflow

### DIFF
--- a/actions/get-changed-files/action.yml
+++ b/actions/get-changed-files/action.yml
@@ -4,6 +4,9 @@ inputs:
   token:
     description: secrets.GITHUB_TOKEN
     required: true
+  directories:
+    description: Prefilter results to only the listed subdirectories (newline-separated)
+    required: false
 outputs:
   files:
     description: A jsonified array of files that were added, modified, or renamed.
@@ -16,7 +19,10 @@ runs:
       with:
         script: |
           try {
-            require('./actions/get-changed-files/index.js')({github, context, core}).catch(err => core.setFailed(err.message))
+            require('./actions/get-changed-files/index.js')({
+              github, context, core,
+              directoriesRaw: `${{ inputs.directories }}`,
+            }).catch(err => core.setFailed(err.message))
           } catch (err) {
             core.setFailed(err.message)
           }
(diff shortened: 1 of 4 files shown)


actions/get-changed-files/action.yml
```diff
@@ -16,7 +19,10 @@ runs:
       with:
         script: |
           try {
-            require('./actions/get-changed-files/index.js')({github, context, core}).catch(err => core.setFailed(err.message))
+            require('./actions/get-changed-files/index.js')({
+              github, context, core,
+              directoriesRaw: `${{ inputs.directories }}`,
```
kevinbarabash, line 24: If `inputs.directories` isn't provided, this will just be an empty string, right?

actions/get-changed-files/index.js
```diff
@@ -18,8 +18,12 @@ const getBaseAndHead = (context, core) => {
     }
 };
 
-module.exports = async ({github, context, core}) => {
-    const absolute = !!core.getInput("absolute", {required: false});
```
kevinbarabash, line 22: I assume this went away to make the filtering based on directory easier.  I'm curious why we were using absolute paths before.

kevinbarabash (approved):
LGTM.  Thanks for adding the ability to pre-filter change files.  I am curious about why it's okay to no longer use absolute paths.

actions/get-changed-files/index.js
jaredly: So we weren't using the `absolute` option anywhere (I had added to interact with a previous version of some actions which we're not using anymore)

actions/get-changed-files/action.yml
jaredly: the default value is `.github/`

jaredly: oh wait jk I was thinking about a different option. Yes, it will be the empty string.

PR #22, opened 2022-06-02
Simpler failure message when a json-args command fails

jaredly:
## Summary:
Currently it's just bubbling up as a javascript error, which is unnecessarily noisy.

Issue: XXX-XXXX

## Test plan:
See node-ci.yml

### DIFF
--- a/.github/workflows/node-ci.yml
+++ b/.github/workflows/node-ci.yml
@@ -79,6 +79,13 @@ jobs:
         id: changed
       - run: echo '${{ steps.changed.outputs.files }}'
 
+      # Test json-args with a failing command
+      - uses: ./actions/json-args
+        continue-on-error: true
+        with:
+          list: '["hi","folks with space"]'
+          run: node -e 'console.log(process.argv); process.exit(1)' {}
+
       - uses: ./actions/check-for-changeset
         with:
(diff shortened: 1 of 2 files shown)


.changeset/modern-starfishes-travel.md
```diff
@@ -0,0 +1,5 @@
+---
```
somewhatabstract, line 1: ASIDE: Disappointed in changesets here. Pretty sure the plural of starfish is starfish, not starfishes.

.github/workflows/node-ci.yml
```diff
@@ -79,6 +79,13 @@ jobs:
         id: changed
       - run: echo '${{ steps.changed.outputs.files }}'
 
+      # Test json-args with a failing command
+      - uses: ./actions/json-args
+        continue-on-error: true
+        with:
+          list: '["hi","folks with space"]'
```
somewhatabstract, line 86: NIT: Could you perhaps make it clear what these are? As in, is the `hi` the command that might run or just an arg? Is the `folks with space` also an arg?

Perhaps something like this may be clearer for future maintenance?

```
list: '["arg1_hi", "arg2_now with spaces"]'
```

somewhatabstract (approved):
Thanks for making this clearer!

LGTM, minor suggestion but nothing blocking.

PR #23, opened 2022-06-02
Remove the need to specify dependencies between actions

jaredly:
## Summary:
I made an error with check-for-changeset, by using another action (`./actions/get-changed-files`) but not listing it in package.json. It worked for local testing, but didn't get translated correctly on build.
This makes it so that all inter-action dependencies get automatically resolved.

Issue: XXX-XXXX

## Test plan:
See the failed version https://github.com/Khan/wonder-blocks/runs/6714028552?check_suite_focus=true
and the succeeding version https://github.com/Khan/wonder-blocks/runs/6714266788?check_suite_focus=true

### DIFF
--- a/.github/workflows/node-ci.yml
+++ b/.github/workflows/node-ci.yml
@@ -79,6 +79,13 @@ jobs:
         id: changed
       - run: echo '${{ steps.changed.outputs.files }}'
 
+      # Test json-args with a failing command
+      - uses: ./actions/json-args
+        continue-on-error: true
+        with:
+          list: '["hi","folks with space"]'
+          run: node -e 'console.log(process.argv); process.exit(1)' {}
+
       - uses: ./actions/check-for-changeset
         with:
           exclude: .github/,utils,.eslintrc.js,.prettierrc.js
--- a/utils/build.mjs
+++ b/utils/build.mjs
@@ -40,10 +40,13 @@ export const processActionYml = (
             to: "${{ github.action_path }}/",
         },
     ];
-    Object.keys(packageJsons).forEach((depName) => {
+    const allActionNames = Object.keys(packageJsons);
+    // For each of the other actions, replace any local-path references to it
+    // to a normal action reference, locked to the current latest version.
+    allActionNames.forEach((actionName) => {
         replacements.push({
-            from: new RegExp(`\\buses: \\./actions/${depName}\\b`, "g"),
-            to: `uses: ${monorepoName}@${depName}-v${packageJsons[depName].version}`,
+            from: new RegExp(`\\buses: \\./actions/${actionName}\\b`, "g"),
+            to: `uses: ${monorepoName}@${actionName}-v${packageJsons[actionName].version}`,
         });
     });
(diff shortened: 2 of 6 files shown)


kevinbarabash:
I don't understand how the changes in this PR allow inter-action deps to be resolved.

utils/build.mjs
```diff
@@ -40,7 +40,7 @@ export const processActionYml = (
             to: "${{ github.action_path }}/",
         },
     ];
-    Object.keys(packageJsons[name].dependencies ?? {}).forEach((depName) => {
+    Object.keys(packageJsons).forEach((depName) => {
```
somewhatabstract, line 43: Is `packageJsons` now a list of dependencies? Or is this just saying that each package is a dependency to all the other packages?

jaredly: `packgeJsons` is the full map of `action name` to `contents of the package.json`

somewhatabstract: So, this code is doing what? I guess I'm trying to understand how the keys of `packageJsons` maps to `depName`. What is this doing exactly?

utils/build.mjs
```diff
@@ -40,10 +40,13 @@ export const processActionYml = (
...
-    Object.keys(packageJsons[name].dependencies ?? {}).forEach((depName) => {
+    const allActionNames = Object.keys(packageJsons);
+    // For each of the other actions, replace any local-path references to it
+    // to a normal action reference, locked to the current latest version.
+    allActionNames.forEach((actionName) => {
         replacements.push({
-            from: new RegExp(`\\buses: \\./actions/${depName}\\b`, "g"),
-            to: `uses: ${monorepoName}@${depName}-v${packageJsons[depName].version}`,
+            from: new RegExp(`\\buses: \\./actions/${actionName}\\b`, "g"),
+            to: `uses: ${monorepoName}@${actionName}-v${packageJsons[actionName].version}`,
```
somewhatabstract, line 49: Oh this is much clearer. Thanks!

somewhatabstract (approved):
Thanks for the clarifying changes. This looks good to me!

.github/workflows/node-ci.yml
```diff
@@ -79,6 +79,13 @@ jobs:
         id: changed
       - run: echo '${{ steps.changed.outputs.files }}'
 
+      # Test json-args with a failing command
+      - uses: ./actions/json-args
+        continue-on-error: true
+        with:
+          list: '["hi","folks with space"]'
+          run: node -e 'console.log(process.argv); process.exit(1)' {}
```
kevinbarabash, line 87: **suggestion (non-blocking):** I forgot that you were using node-ci.yml for testing.  You may want to rename the workflow file to something like internal-action-tests.yml (or something like that) so that it's more obvious that changes like these are meant to be committed (I initially thought this was temporary test code that should be removed).

kevinbarabash (approved):
LGTM 🎉

jaredly:
I hadn't realized that this change actually breaks changeset :( instead the thing to do is "lint against using another action w/o listing it in the package.json"

PR #25, opened 2022-06-07
Better error reporting when actions are misconfigured

jaredly:
## Summary:
Currently it says "SyntaxError: Unexpected end of JSON input", which isn't terribly informative. https://github.com/Khan/wonder-stuff/runs/6781773606?check_suite_focus=true

Issue: XXX-XXXX

## Test plan:
see check


jeresig (approved):
Ah - that is better! Thank you

PR #27, opened 2022-06-09
Skip gerald steps if the PR is a draft

jaredly:
## Summary:
We /could/ fix this by putting an `if` on all of the gerald workflows in all of the repos, but this will fix it centrally for all.

Issue: XXX-XXXX

## Test plan:
Make a draft PR off of this one, see that the "gerald" step is skipped, and doesn't auto-assign reviewers
https://github.com/Khan/actions/pull/28 yay


jeresig (approved):
Great, thank you!

PR #33, opened 2022-06-10
Make setup-keeper action less flaky

jaredly:
## Summary:
For some inscrutable reason, using the bash HEREDOC stuff worked a couple of times, and then never again. I trust javascript more anyway, so I'm happy with this change.

## Test plan:
see the check working!

### DIFF
--- a/actions/setup-keeper/action.yml
+++ b/actions/setup-keeper/action.yml
@@ -15,7 +15,7 @@ runs:
       with:
         script:
           const config = `${{ inputs.config }}`;
-          require('fs').writeFileSync('./.keeper-config.json', config)
+          require('fs').writeFileSync(process.env.HOME + '/.keeper-config.json', config)
 
     - name: Set up the keeper config file
       shell: bash
--- a/actions/setup-keeper/action.yml
+++ b/actions/setup-keeper/action.yml
@@ -11,16 +11,16 @@ runs:
       shell: bash
       run: pip3 install cryptography==37.0.2 pyOpenSSL==22.0.0 keepercommander==16.6.5
 
+    - uses: actions/github-script@v6
+      with:
+        script:
+          const config = `${{ inputs.config }}`;
+          require('fs').writeFileSync(process.env.HOME + '/.keeper-config.json', config)
+
     - name: Set up the keeper config file
       shell: bash
-      run: |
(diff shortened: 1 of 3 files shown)


actions/setup-keeper/action.yml
```diff
@@ -11,14 +11,16 @@ runs:
       shell: bash
       run: pip3 install cryptography==37.0.2 pyOpenSSL==22.0.0 keepercommander==16.6.5
 
+    - uses: actions/github-script@v6
+      with:
+        script:
+          const config = `${{ inputs.config }}`;
```
jeresig, line 17: I'm surprised there isn't a way to access incoming inputs as variables in `github-script`?

actions/setup-keeper/action.yml
```diff
@@ -11,14 +11,16 @@ runs:
...
-        cat << EOF > $HOME/.keeper-config.json
-        ${{ inputs.config }}
-        EOF
-        echo "KEEPER_CONFIG_FILE=$HOME/.keeper-config.json" >> $GITHUB_ENV
+      run: echo "KEEPER_CONFIG_FILE=$HOME/.keeper-config.json" >> $GITHUB_ENV
 
     - name: Check that keeper is installed correctly
       shell: bash
-      run: keeper whoami | grep '@khanacademy.org'
+      run: keeper this-device
```
jeresig, line 26: What is this change doing?

jeresig (approved):
Huh - I wonder why it was failing with the other technique?

actions/setup-keeper/action.yml
jaredly: making the consistency check more generic (not relying on the ka.org namespace). Either command will fail if authentication didn't happen correctly.

actions/setup-keeper/action.yml
jaredly: yeah I was also surprised, but composite actions don't get the ENV variables made for inputs automatically. I guess it makes usage more explicit?

jeresig: I wonder if you could use this: https://github.com/actions/github-script#use-env-as-input

so define an ENV for the config for this step and then access it via `process.env.KEEPER_CONFIG`.

PR #34, opened 2022-11-22
Update to use latest actions, removing Node12 dep

somewhatabstract:
## Summary:
This updates to use the latest actions and therefore remove the Node 12 dependency that is soon to not even be supported by GitHub Actions.

Issue: FEI-4813

## Test plan:
Put this up, check things are OK. Then update and check the things that use this thing.


jeresig (approved):
Nice - thanks for upgrading all of this!

PR #43, opened 2023-05-12
Allow check-for-changeset to be configured to exclude files by extension

jeremywiebe:
## Summary:

The check-for-changeset action is really useful in making sure we don't put up a PR without a `.changeset/*` file. But it often flags PRs that we can guarantee don't need a changeset (for example, a test-only PR or one that just touches Storybook stories). 

This PR adds the ability to filter out entire classes of files by extension (*.test.js or *.stories.js) for example. 

We will still need to configure each repo that uses this action to exclude the files by extension, but this opens the door for those configurations.

Issue: "none"

## Test plan:

Use the action in this PR in a repo (Khan/perseus, for example) and create a PR that contains changes only to files with extensions we configure to exclude.

### DIFF
--- a/actions/check-for-changeset/action.yml
+++ b/actions/check-for-changeset/action.yml
@@ -9,6 +9,10 @@ inputs:
     description: a list of comma-separated extensions that don't need a changeset when changed. Default <empty>
     required: false
     default: ''
+  exclude_glob:
+    description: a list of comma-separated globs (using picomatch syntax) that don't need a changeset when changed. Default <empty>
+    required: false
+    default: ''
 
 runs:
   using: "composite"
@@ -23,7 +27,7 @@ runs:
         invert: true
         files: ${{ inputs.exclude }}
         extensions: $${{ inputs.exclude_extensions }}
-        file_globs: ${{ inputs.exclude_glob }}
+        globs: ${{ inputs.exclude_glob }}
 
     - uses: actions/github-script@v6
       with:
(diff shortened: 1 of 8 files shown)


jaredly (approved):
love it! Yeah, test files and like `.md` files would be great to exclude

actions/check-for-changeset/action.yml
```diff
@@ -5,6 +5,10 @@ inputs:
     description: list of comma-separated paths (files or directories ending in /) that don't need a changeset when changed. Default .github/
     required: false
     default: .github/
+  exclude_extensions:
+    description: a list of comma-separated extensions that don't need a changeset when changed. Default <empty>
+    required: false
+    default: ''
```
jaredly, line 11: oh you forgot `exclude_glob` here too I think?

jeremywiebe: Oops... I left `exclude_glob` in by accident.  I saw we don't currently have any sort of `glob` dependency and was wondering if that'd be ok to add.  If so, I'll add it and plumb support through.

actions/filter-files/index.js
```diff
@@ -11,15 +13,24 @@ const parseList = (raw) => {
...
+    extensionsRaw,
+    exactFilesRaw,
+    globsRaw,
+    inputFiles,
+    invert,
+    core,
+}) => {
     const extensions = parseList(extensionsRaw);
     const exactFiles = parseList(exactFilesRaw);
+    const globMather = picomatch(parseList(globsRaw));
```
jeremywiebe, line 26: `picomatch` can handle an array of glob expressions, and an empty array doesn't match anything.

jaredly (approved):
cool!

PR #44, opened 2023-05-12
Support JS-based actions having dependencies

jeremywiebe:
## Summary:

In the last Version Packages PR I had a change that added a npm dependency to one of the JS-based actions in this repo and that [broke](https://github.com/Khan/actions/actions/runs/4963419816/jobs/8882640617) the publish step!  We currently don't support this and so the publish step failed for (for two reasons actually):

  1. Our build script assumes that all `dependencies` in our actions `package.json` are a cross-reference to another action in this repo
  2. We don't have any bundling setup so our JS actions are copied as-is into the `dist` folder that we publish

This PR addresses both of these issues by:
  1. We don't try to process depedencies in action YML files for external deps
  2. Set up ncc to build JS packages on publish (following Github [JS action template](https://github.com/actions/javascript-action/blob/main/package.json#L8))

Issue: "none"

## Test plan:

I've tested the build using `node utils/run-publish.mjs --dry-run` and it seems to build the files into the `dist` folders correctly.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -26,4 +26,12 @@ module.exports = {
         sourceType: "module",
         ecmaVersion: 2022,
     },
+    overrides: [
+        {
+            files: "utils/*",
+            rules: {
+                "no-console": "off",
+            },
+        },
+    ],
 };
--- a/utils/build.mjs
+++ b/utils/build.mjs
@@ -5,6 +5,7 @@
  */
 import fs from "fs";
 import path from "path";
+import {execSync} from "child_process";
 
 function copyDir(src, dest) {
     const entries = fs.readdirSync(src, {withFileTypes: true});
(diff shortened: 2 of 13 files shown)


.eslintrc.js
```diff
@@ -26,4 +26,12 @@ module.exports = {
         sourceType: "module",
         ecmaVersion: 2022,
     },
+    overrides: [
+        {
+            files: "utils/*",
+            rules: {
+                "no-console": "off",
+            },
```
jeremywiebe, line 34: In our utils, I think it's useful to be able to log output to the console for local dev but also so when the action has issues we have more info.

somewhatabstract:
Is this what broke the changesets check too?

https://github.com/somewhatabstract/checksync/actions/runs/4975521186/jobs/8902761233?pr=1271

How did it break an already tagged release?

utils/build.mjs
```diff
@@ -65,5 +71,17 @@ export const buildPackage = (name, packageJsons, monorepoName) => {
...
         processActionYml(name, packageJsons, actionYml, monorepoName),
     );
+
+    // Build the JS to the dist folder (so we can support using npm deps)
+    // We need to delete the index file first because `copyDir` would have
+    // copied it to the `dist` folder already but `ncc` won't overwrite an
+    // existing output file.
+    if (fs.existsSync(`${dist}/index.js`)) {
+        fs.rmSync(`${dist}/index.js`, {force: true});
+        execSync(
```
jeresig, line 81: It seems like we only want to rm it if it exists - but we should be running the ncc command even if the file doesn't exist, right?

jeresig (approved):
Thank you for fixing this up!

jaredly (approved):
It does seem like a little bit of a "gotcha" that in order for `ncc` to run, there needs to be an `index.js` 🤔 could you add some validation in `build.mjs` that verifies that, if there's an external dependency, there also exists an `index.js`?

utils/build.mjs
jeremywiebe: Fixed.  We now delete the build file only if it exists, but then only build if a source file exists.

PR #46, opened 2023-05-15
Define in-repo deps for check-for-changeset

jeremywiebe:
## Summary:

In the last release, I added some build setup so that our actions in this repo could have npm dependencies. This was done using `ncc` to build the `index.js` for each acttion and bundle all dependencies into the dist/index.js file. 

When I integrated this release though, the `check-for-changeset` action [failed](https://github.com/Khan/perseus/actions/runs/4983846372/jobs/8921407849?pr=530) with this message: `Error: Can't find 'action.yml', 'action.yaml' or 'Dockerfile' under '/home/runner/work/perseus/perseus/actions/get-changed-files'. Did you forget to run actions/checkout before running your local action?`

image

I checked this repo and it looks like the build expects any action dependencies to be defined in `package.json` of the action (for `check-for-changeset` it depends on `get-changed-files` and `filter-files` and neither of these were listed in the `package.json`.

Issue: "none"

## Test plan:

Publish and test in a repo that uses this action.


jeresig (approved):
Ahh!! Good catch.

PR #69, opened 2023-08-29
Remove the keeper-related github actions.

csilvers:
## Summary:
We're moving to using GSM for secrets management, at least for secrets
used in code.  So github actions no longer need to be able to access
keeper.

Issue: https://khanacademy.atlassian.net/browse/INFRA-9448

## Test plan:
None


jeresig (approved):
Great - thanks for cleaning this up!

PR #72, opened 2023-09-21
Remove superfluous input

nedredmond:
`token` input is unused, but marked `required`. This creates some noisy red underlines in consuming workflows that use the Github Actions extension.

Screenshot 2023-09-21 at 4 52 14 PM


jeresig (approved):

jeresig:
Thanks for the fix, Ned!

PR #74, opened 2023-10-02
Allow Conjunctive Filter (all true, vs current any true)

nedredmond:
Adds option to assert _all_ filters pass.

Currently, if any of the filters are true, the path is included in filtered files. This makes it difficult to use multiple filter types to refine our choice, especially when not in `invert` mode.

Now, we can designate some directories and file extensions and only files in those directories with those extensions will return.

### DIFF
--- a/actions/filter-files/action.yml
+++ b/actions/filter-files/action.yml
@@ -1,46 +1,46 @@
-name: "Filter files please"
-description: "Filter the list of changed files"
+name: 'Filter files please'
+description: 'Filter the list of changed files'
 inputs:
-    changed-files:
-        description: "jsonified list of changed files"
-        required: true
-    files:
-        description: 'comma- or newline-separated list of files to match against. names ending in "/" will be treated as a directory path, and match as a prefix. Otherwise path names must be a full match.'
-        required: false
-    extensions:
-        description: "comma- or newline-separated list of extensions to check for"
-        required: false
-    globs:
-        description: "comma- or newline-separated list of globs (using picomatch syntax) to check for"
-        required: false
-    invert:
-        description: "if true, return the non-matched paths instead of the matched ones."
-        required: false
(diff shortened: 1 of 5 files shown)


actions/filter-files/index.js
```diff
@@ -64,18 +64,41 @@ module.exports = ({
...
     invert,
+    conjunctive,
     core,
 }) => {
-    const extensions = parseList(extensionsRaw);
-    const exactFiles = parseList(exactFilesRaw);
-    const globMatcher = picomatch(parseList(globsRaw));
-    const directories = exactFiles.filter((name) => name.endsWith("/"));
+    const filters = [];
+    if (exactFilesRaw) {
```
nedredmond, line 71: In order for `every` to work, we should only include filters that have been specified.

actions/filter-files/index.js
```diff
@@ -64,18 +64,41 @@ module.exports = ({
...
+        const paths = parseList(exactFilesRaw);
+        const [directories, exactFiles] = paths.reduce(
```
nedredmond, line 73: I've partitioned this rather than leave redundant entries in `exactFiles`.

actions/filter-files/action.yml
```diff
@@ -35,7 +40,8 @@ runs:
             throw new Error(`filter-files was called with an empty string as the "changed-files" parameter.`)
           }
           const inputFiles = JSON.parse(inputFilesRaw);
-          const invert = `${{ inputs.invert }}` == 'true';
+          const invert = ${{ inputs.invert == 'true' }};
```
nedredmond, line 43: Instead of logging `invert = 'true' == 'true'` it now logs `invert = true`
Instead of `invert = 'true' == 'false'` now logs `invert = false`

jaredly (approved):
looks good!

PR #76, opened 2023-10-04
Add matchAllGlobs to allow conjunctive glob matching

nedredmond:
Adds ability to evaluate globs in conjunctive manner as well


jaredly (approved):

jandrade (approved):
Looks great, thanks for adding this!

PR #83, opened 2024-03-26
Update to Node20-ready action versions

somewhatabstract:
## Summary:
This updates all the workflows to use actions that are running on Node 20 instead of 16.

Issue: FEI-5484

## Test plan:
`yarn test`


jeresig (approved):
Great, thank you Jeff!

PR #85, opened 2024-04-05
Update gerald action to Node 20

somewhatabstract:
## Summary:
This updates the `gerald-pr` action to use the latest Khan/gerald release, supporting Node 20, and the repo version of `get-changed-files` so that it will be up-to-date on each release.

Issue: FEI-5545

## Test plan:
`yarn test`
Update consumers of this action to use the new version and see that they no longer warn of Node 16 deprecation.

### DIFF
--- a/actions/gerald-pr/action.yml
+++ b/actions/gerald-pr/action.yml
@@ -11,7 +11,7 @@ runs:
   using: "composite"
   steps:
       - name: Get All Changed Files
-        uses: Khan/actions@get-changed-files-v1
+        uses: ./actions/get-changed-files
         id: changed
         if: ${{ github.event.pull_request.draft == false }}
       - name: Check out base branch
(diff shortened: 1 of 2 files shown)


actions/gerald-pr/action.yml
```diff
@@ -11,7 +11,7 @@ runs:
   using: "composite"
   steps:
       - name: Get All Changed Files
-        uses: Khan/actions@get-changed-files-v1
+        uses: ./actions/get-changed-files
```
somewhatabstract, line 14: **note:** This seems like a more appropriate way to reference this action. With any given release we would want to use the version of the action that was included in that release.

actions/gerald-pr/action.yml
```diff
@@ -25,7 +25,7 @@ runs:
           ref: '${{ github.head_ref }}'
         if: ${{ github.event.pull_request.draft == false }}
       - name: Run Gerald
-        uses: Khan/gerald@v3.4
+        uses: Khan/gerald@v4.0
```
somewhatabstract, line 28: **note:** Updating to the Node 20 version of Gerald.

jeresig (approved):
Thank you!

PR #91, opened 2024-06-04
Bumping actions/gerald-pr to point to newer version of Khan/gerald

lillialexis:
## Summary:
See [this PR](https://github.com/Khan/gerald/pull/87) with a bug-fix regarding GitHub teams not getting added when author is a member of that team, and the author adds comments. Point this to that, once it's landed and tagged.

I also updated the README with instructions (mostly for myself, because I forgot what I did last time) on how to test PRs in this repo _before_ landing (and publishing).

Issue: XXX-XXXX

## Test plan:

### DIFF
--- a/actions/gerald-pr/action.yml
+++ b/actions/gerald-pr/action.yml
@@ -25,7 +25,7 @@ runs:
           ref: '${{ github.head_ref }}'
         if: ${{ github.event.pull_request.draft == false }}
       - name: Run Gerald
-        uses: Khan/gerald@v4
+        uses: Khan/gerald@v4.1
         env:
           GITHUB_TOKEN: '${{ inputs.token }}'
           ADMIN_PERMISSION_TOKEN: '${{ inputs.admin-token }}'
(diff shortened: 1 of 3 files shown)


actions/gerald-pr/action.yml
```diff
@@ -25,7 +25,7 @@ runs:
           ref: '${{ github.head_ref }}'
         if: ${{ github.event.pull_request.draft == false }}
       - name: Run Gerald
-        uses: Khan/gerald@v4.0
+        uses: Khan/gerald@v4.1
```
jeresig, line 28: It may be better to just change this to `v4` so that minor updates will get rolled out automatically.

jeresig (approved):

actions/gerald-pr/action.yml
lillialexis: ooooh... I didn't know it can do that! That's interesting... I would have thought that anything after the `@` was just an opaque git ref. What part of the system understands semantic versioning embedded in git tags?

jeresig: Ah - you know what, I take it back! So this only works for things published in this repo. Gerald isn't set up to do this yet. We have to hardcode the version here... ☹️ If actions here are published you can reference them by major version, which is nice!

PR #99, opened 2025-03-19
Allow running shared-node-cache without ssh-private-key

jeremywiebe:
## Summary:

In the last PR I added a `ssh-private-key` input to the `shared-node-cache` action. However, this isn't needed/used in all cases and so making it required breaks some existing usages. I've made it optional now and only run the `webfactory/ssh-agent` action if a key has been provided. 

Issue: --none--

## Test plan:

Watch the actions. The `shared-node-cache` action is run in this repo as part of PR checks. That run is now set up to _not_ provide an `ssh-private-key`. I'll verify that it works _with_ an `ssh-private-key` after landing and integrating in another repo.

### DIFF
--- a/.changeset/nervous-actors-tell.md
+++ b/.changeset/nervous-actors-tell.md
@@ -1,0 +1,5 @@
+---
+"shared-node-cache": minor
+---
+
+Make `ssh-private-key` optional on the shared-node-cache action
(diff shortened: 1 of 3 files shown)


jeresig (approved):
Ah yeah, thanks Jeremy!

PR #102, opened 2025-05-19
Change behavior of "negative globs" to match .gitignore

jaredly:
## Summary:
The current behavior of negative globs only makes sense if `matchAllGlobs` is true. If it's false, then weird things happen. For example:
```
input files:
- one/two.js
- one/two/three.test.ts
- one/two.json
globs:
one/**
!**/*.test.ts
!**/*.json
```
The behavior is "a file 'matches' if any of the globs match". As you can see in this case, given two negative non-intersecting globs, all files will always match, because at least one of them will match for every file.

I'm changing it so that if `matchAllGlobs` is false, then the behavior is: "all negative globs have to match, and at least one positive glob needs to match".
With this, the result of the above example would be `one/two.js`, which I think is intuitive.

Issue: none

## Test plan:
See the test I added.

### DIFF
--- a/actions/filter-files/filter-files.test.mjs
+++ b/actions/filter-files/filter-files.test.mjs
@@ -13,7 +13,6 @@ describe("filterFiles", () => {
         const result = filterFiles({
             globsRaw: `packages/*\n!packages/*.json`,
             inputFiles,
-            matchAllGlobs: true,
             core,
         });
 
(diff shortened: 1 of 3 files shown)


actions/filter-files/filter-files.test.mjs
```diff
@@ -13,7 +13,6 @@ describe("filterFiles", () => {
         const result = filterFiles({
             globsRaw: `packages/*\n!packages/*.json`,
             inputFiles,
-            matchAllGlobs: true,
```
jaredly, line 16: before I had to set `matchAllGlobs` to true in order for the "NOT glob" to be usefully applied.

PR #171, opened 2026-04-09
Add in more secure-network logging.

jeresig:
## Summary:
This adds in a bunch more logging info, including one as a post step, to help us understand what's happening with upstash and if it's still running, of if it's crashing, or something else. Especially in the context of Github's hosted runners.

Issue: FEI-XXXX

## Test plan:

### DIFF
--- a/actions/secure-network/action.yml
+++ b/actions/secure-network/action.yml
@@ -1,8 +1,8 @@
 name: "Secure Network"
 description: >-
-  Set up a DNS firewall that restricts outbound DNS to an explicit allowlist.
-  Core GitHub domains are always included. Pass additional .conf files and/or
-  extra domains for any other hosts your workflow needs to reach.
+    Set up a DNS firewall that restricts outbound DNS to an explicit allowlist.
+    Core GitHub domains are always included. Pass additional .conf files and/or
+    extra domains for any other hosts your workflow needs to reach.
 
 inputs:
     conf-files:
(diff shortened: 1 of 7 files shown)


actions/secure-network/action.yml
```diff
@@ -19,18 +19,7 @@ inputs:
...
-          run: |
-              # Note: secure-network.js is invoked with node directly rather than via
-              # actions/github-script or a TypeScript entrypoint because it is a
-              # large, security-critical script with its own bundled dependencies
-              # (github.conf, github-meta.json). The actions/github-script approach
-              # is designed for small inline scripts, not multi-file bundles like this.
-              sudo env "PATH=$PATH" "CONF_FILES=$CONF_FILES" "EXTRA_DOMAINS=$EXTRA_DOMAINS" \
-                node "${{ github.action_path }}/secure-network.js"
+    using: "node24"
+    main: "main.js"
```
jaredly, line 23: ooh fancy

actions/secure-network/action.yml
```diff
@@ -19,18 +19,7 @@ inputs:
...
+    post: "post.js"
+    post-if: "always()"
```
jaredly: > If not set, then post-if defaults to always()

https://docs.github.com/en/actions/reference/workflows-and-actions/metadata-syntax#runspost-if

jaredly (approved):
solid

actions/secure-network/action.yml
jeresig: Oh good catch!

PR #173, opened 2026-04-13
fix-workflows: exempt macos- runners from all fixups

jeresig:
Jobs whose \`runs-on\` starts with \`macos-\` are now skipped entirely by the
fix-workflows action — no secure-network step insertion, no runs-on
rewrite, and no violation reported in the verification pass.

Introduces \`isExemptRunner\` as the single source of truth for the
exemption, and extracts \`processJob\` / \`checkJob\` so the per-job logic is
unit-testable without filesystem I/O (37 tests, all passing).

Reviewers: #frontend-infra-web

## Test plan:

All behaviour is covered by unit tests — no manual steps required.

## Review plan:

No High or Medium risk file changes. The only production-code change is a
focused guard clause in \`actions/fix-workflows/index.ts\` backed by full
test coverage.

### DIFF
--- a/actions/fix-workflows/index.ts
+++ b/actions/fix-workflows/index.ts
@@ -209,7 +209,10 @@ type JobOptions = {
 export function processJob(
     doc: Document,
     job: YAMLMap,
-    {shouldFixRunsOn = false, setupAction = DEFAULT_SETUP_ACTION}: JobOptions = {},
+    {
+        shouldFixRunsOn = false,
+        setupAction = DEFAULT_SETUP_ACTION,
+    }: JobOptions = {},
 ): boolean {
     if (isExemptRunner(job)) {
         return false;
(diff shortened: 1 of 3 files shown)


actions/fix-workflows/index.ts
```diff
@@ -185,38 +197,79 @@ export function fixRunsOn(job: YAMLMap): boolean {
...
+    }: JobOptions = {},
+): boolean {
+    if (isExemptRunner(job)) {
+        return false;
+    }
+    let changed = false;
+    if (shouldFixRunsOn) {
+        changed = fixRunsOn(job) || changed;
+    }
+    const steps = (job as any).get("steps");
```
jaredly, line 224: I'm surprised by these `as any`s. YAMLMap includes a .get (used above in isExpemptRunner). Is the return value wrong?

jaredly (approved):
had a question,  but overall looks good!

actions/fix-workflows/index.ts
jeresig: It looks like in YAML-land it doesn't know what the actual object could be - it could be a string, a number, an object, an array, etc. - it looks like all return values are actually marked as "unknown" so in this case it's just casting it to any so that it can access the expected value (this would error out if something else besides an object was here, but that's probably to be expected).

PR #176, opened 2026-05-12
generate-terraform-plan: clarify scheduled PR body about image digest churn

jwbron:
## Summary
- The "Scheduled Build Context" section of the auto-generated terraform-plan PR previously told reviewers to dig through Cloud Build logs whenever the plan showed only a Docker image digest change. In practice that case is almost always just an upstream republish of the base image (e.g. `python:3.11-slim`) or new versions of OS packages pulled in by the Dockerfile — the daily rebuild produces a new digest even though nothing in the repo changed.
- Replace the cloud-build-logs walkthrough with a one-sentence explanation that this case is expected and safe to merge.

Context: see [Khan/internal-services#514](https://github.com/Khan/internal-services/pull/514) for an example of a PR that triggered this confusion. The digest there changed because a new `python:3.11-slim` was republished upstream, which invalidated the Dockerfile cache and caused a fresh `apt-get install` against Debian trixie.

## Test plan
- [ ] Tag a new patch release of `generate-terraform-plan` (changeset included)
- [ ] Bump the action reference in `Khan/internal-services` `.github/workflows/build-github-actions-runner-terraform-plan.yml` to the new tag
- [ ] Wait for the next scheduled (cron) run and confirm the PR body shows the new wording

### DIFF
--- a/.changeset/clarify-scheduled-plan-pr-body.md
+++ b/.changeset/clarify-scheduled-plan-pr-body.md
@@ -1,0 +1,5 @@
+---
+"generate-terraform-plan": patch
+---
+
+Clarify the scheduled-plan PR description: when the diff shows no code changes but the plan shows a Docker image digest change, explain that this is from an upstream base-image or OS-package update and is safe to merge. Replaces the previous (and confusing) instructions to dig through Cloud Build logs.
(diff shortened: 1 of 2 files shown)


csilvers (approved):
Thank you!

PR #185, opened 2026-05-28
fix-workflows: replace auto-commit with local CLI for fixing and formatting

jeresig:
## Summary

- **Stop auto-committing** fixed workflow files in CI. Instead, the action now shows the developer a single `pnpm dlx` command to run locally.
- **Add CLI entrypoint** (`cli.ts`) that runs both the lint fixer and oxfmt formatter in one step, so devs only need one command.
- **Skip `*.lock.yml` files** during workflow processing to avoid modifying lock files.
- **Derive version tag dynamically** from `package.json` so the printed `pnpm dlx` command always references the correct published tag.

## What changed

| File | Change |
|------|--------|
| `actions/fix-workflows/action.yml` | Replaced "Commit changes" step with "Show fix instructions" that prints a `pnpm dlx` command and lists files that need updating |
| `actions/fix-workflows/cli.ts` | New CLI entrypoint — parses `--fix-runs-on` / `--setup-action` args, runs fixer, then spawns oxfmt |
| `actions/fix-workflows/index.ts` | Added `*.lock.yml` exclusion in `getFilesToCheck()` |
| `actions/fix-workflows/package.json` | Added `bin` field pointing to `cli.js`; added `cli.js` to `files` array for publishing |
| `actions/tsconfig.json` | Added `fix-workflows/cli.ts` to `include` |
| `.changeset/fix-workflows-no-commit.md` | Minor changeset |

## How it works in CI now

1. Action fixes workflow violations (same as before)
2. Action formats with oxfmt (same as before)  
3. If files changed, prints a warning with the exact command:
   ```
   pnpm dlx github:Khan/actions#fix-workflows-v3 --setup-action 'Khan/actions@secure-network-v1'
   ```
4. Lists the files that need to be updated

## Test plan

- [x] All 159 tests pass across 8 test files
- [x] `pnpm run build` succeeds (including new `cli.ts`)
- [x] CLI runs successfully locally (`node actions/fix-workflows/cli.js`)
- [ ] Verify `pnpm dlx github:Khan/actions#fix-workflows-v3` works after publishing

### DIFF
--- a/actions/fix-workflows/action.yml
+++ b/actions/fix-workflows/action.yml
@@ -51,11 +51,8 @@ runs:
                 MAJOR_VERSION=$(node -p "require('${{ github.action_path }}/package.json').version.split('.')[0]")
                 TAG="fix-workflows-v${MAJOR_VERSION}"
                 ARGS="${{ inputs.fix-runs-on == 'true' && ' --fix-runs-on' || '' }} --setup-action '${{ inputs.setup-action }}'"
-                echo ""
-                echo "::warning::Workflow files need updating. Run the following command locally to fix and format them:"
-                echo ""
-                echo "  pnpm dlx github:Khan/actions#${TAG}${ARGS}"
-                echo ""
-                echo "Files that need to be updated:"
-                git diff --name-only .github/workflows/
+                CMD="pnpm dlx github:Khan/actions#${TAG}${ARGS}"
+                for file in $(git diff --name-only .github/workflows/); do
+                  echo "::warning file=${file}::This file needs updating. Run: ${CMD}"
+                done
               fi
--- a/actions/fix-workflows/action.yml
+++ b/actions/fix-workflows/action.yml
@@ -42,15 +42,20 @@ runs:
                 --config "${{ github.action_path }}/.oxfmtrc.json" \
                 ".github/workflows/**/*.yml"
 
-        - name: Commit changes
+        - name: Show fix instructions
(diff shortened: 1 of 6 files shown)


actions/fix-workflows/action.yml
```diff
@@ -42,15 +42,20 @@ runs:
...
+                MAJOR_VERSION=$(node -p "require('${{ github.action_path }}/package.json').version.split('.')[0]")
+                TAG="fix-workflows-v${MAJOR_VERSION}"
+                ARGS="${{ inputs.fix-runs-on == 'true' && ' --fix-runs-on' || '' }} --setup-action '${{ inputs.setup-action }}'"
+                echo ""
+                echo "::warning::Workflow files need updating. Run the following command locally to fix and format them:"
+                echo ""
+                echo "  pnpm dlx github:Khan/actions#${TAG}${ARGS}"
+                echo ""
+                echo "Files that need to be updated:"
+                git diff --name-only .github/workflows/
```
jaredly: can we have this be an inline review comment / lint error, instead of just in the actions run log?

jaredly (approved):
looks reasonable

actions/fix-workflows/action.yml
jeresig: I think the `::warning::` makes it an inline error in the workflow - but yeah, I'll see if it can add it as an annotation on each of the files that need to be updated.

PR #189, opened 2026-06-17
Add shared review PR-reviewer workflow + workflow publishing

jeresig:
Adds `workflows/review/` — a generic, reusable gh-aw PR-reviewer workflow installable via `gh aw add Khan/actions/workflows/review/review.md`. The flow is repo-agnostic (frontend **or** backend); consumers supply their own risk patterns, best-practice skills, CI-tooling exclusions, and reviewer team allowlist through gh-aw imports under their own `.github/aw/review/`. Frontend consumer PR: Khan/frontend#12614.

### What's here
- `workflows/review/review.md` — the shared workflow. Language-neutral prose/examples; egress `[defaults, github, *.sentry.io]`. Uses `{{#runtime-import? aw/review/*.md}}` for the consumer's risk/skills/ci-tooling snippets and `imports: .github/aw/review/config.md` for the consumer's `add-reviewer` (its team allowlist is repo-specific). Khan-wide infra is baked in directly: the OTLP → Sentry observability config, the `deploy/*` + `changeset-release/main` `if:` skip, and `resolve-pull-request-review-thread` (with the `KHAN_ACTIONS_BOT_TOKEN`). `add-reviewer` is intentionally **not** defined here (gh-aw lets the main workflow override an imported safe-output, which would discard the consumer's allowlist).
- `workflows/review/{package.json, README.md}` + a **major** changeset → first release `review-v1.0.0`.
- Publish tooling: `publishWorkflowsAsNeeded` (`utils/publish.ts` → `run-publish.ts`) tags workflow packages on the **real commit tree** (so `workflows/<name>/<file>.md@<ref>` resolves for `gh aw add`), distinct from the tree-rewritten bare tags `actions/` use. Factored into module-level `publishWorkflow()` / `runCommands()` helpers. `workflows/*` added to `pnpm-workspace.yaml`.

### Reviewer behavior
- Approving reviews leave **no message** (empty body, no footer).
- The workflow **resolves its own prior threads** once their issue is addressed (`resolve-pull-request-review-thread`, using `KHAN_ACTIONS_BOT_TOKEN` — the default GITHUB_TOKEN can hit "Resource not accessible by integration" on bot-authored threads) instead of replying — reply-to removed.
- Comments are steered toward **concision** (one or two sentences, no preamble).
[...]

PR #190, opened 2026-06-25
review: delegate PR review to inline sub-agents

jeresig:
## What

Restructures the shared `review` PR-reviewer agentic workflow (added in #189) to run its analysis through inline gh-aw **sub-agents** instead of one monolithic agent.

**Phase 1 — triage.** `pattern-triage` finds common cross-file patterns and returns the set of files that actually need review — dropping generated, formatting-only, and pattern-only changes. The orchestrator then narrows the staged diff/file-list to that set.

**Phase 2 — parallel review** (on the narrowed set):
- `correctness-reviewer` — per-file risk + correctness findings
- `skill-auditor` — best-practice skill violations
- `reviewer-mapper` — file → owning team (from `.github/REVIEWERS`)
- `thread-reconciler` — which earlier bot threads are now addressed

Reviewers run on lower-cost models (`small`/`large` aliases) and read the diff + `.github/aw/review/` config from the checked-out repo. The main agent makes every GitHub and safe-output call — sub-agents have neither.

## Other changes
- Gather context before the early-exit check (fingerprint computed once, in gather).
- Fix the empty-body APPROVE rejection (explicit no-inline-comment path).
- Read repo files from the checkout instead of the GitHub API; drop the `AGENTS.md` read.
- Fingerprint the diff by per-file patch hash (the GitHub MCP exposes no blob `sha`).
- The repo-specific `risk-classification.md` / `ci-tooling.md` / `skills.md` imports are now **required**.

## Notes for reviewers
- Minor changeset included → `review` 1.0.0 → 1.1.0.
- Motivated by analyzing several frontend `review` runs: reduce per-run context bloat and cost, and parallelize the bulk of the work.
- ⚠️ Not yet validated with `gh aw compile` — please run a compile check before merge.


jeresig:
@claude review once

workflows/review/review.md
```diff
@@ -633,3 +588,159 @@ Save to `/tmp/gh-aw/cache-memory/pr-${{ github.event.pull_request.number || gith
...
+
+Return ONLY this JSON object (no prose, no code fence):
+{
+  "owners": {"path/to/file": ["team-a", "team-b"], "path/with/no/owner": []}
+}
+
+## agent: `thread-reconciler`
+---
+description: Decides which of the workflow's earlier review threads the current code has addressed; returns thread ids.
+model: small
```
jwbron: I would switch this to `large` (or opus). This one is not a simple, mechanical review. For each earlier bot thread it has to (a) understand what a prior free-text review comment was actually complaining about, then (b) read the current file and judge whether that concern is now fixed, removed, or moot. This is a semantic comprehension + judgment call, which haiku does not excel at. A false "resolve" silently closes a still-valid blocking thread, hiding an unaddressed issue from humans. The prompt's "when in doubt, keep it" bias mitigates this, but leaning on prompt discipline to compensate for an under-powered model is fragile. The cost delta of using a stronger model here is small relative to the risk of suppressing a real finding.

workflows/review/review.md
```diff
@@ -633,3 +588,159 @@ Save to `/tmp/gh-aw/cache-memory/pr-${{ github.event.pull_request.number || gith
 - No sarcasm, condescension, or excessive praise.
 - No emoji in comments.
 - Comment on code, not people. Critique the work, not the author.
+
+## agent: `correctness-reviewer`
+---
+description: Classifies each changed file's risk and reviews the diff for correctness defects; returns JSON.
+model: large
```
jwbron: from what I can tell, this maps to sonnet right? I don't think that's the appropriate model to use for what's effectively the general review bot. Opus reasoning is significantly better and will catch more issues + make fewer false claims. We might even consider bumping other reviewers here up to opus as well.

workflows/review/review.md
```diff
@@ -132,152 +132,97 @@ helpful. State facts, not opinions about code taste.
...
+  blocking). Use them for the verdict (Step 4) and the inline comments (Step 5).
+- **`reviewer-mapper`** — returns `owners` (`{path: [team, …]}`). Steps 7 and 8 use
+  this mapping directly.
+- **`thread-reconciler`** — returns `{resolve: [...], keep: [...]}` over the threads
+  you staged. Resolve each `thread_id` in `resolve` with the
+  `resolve-pull-request-review-thread` safe output (yours to do — sub-agents cannot);
+  never reply to a thread, and for a `keep` thread do not open a duplicate comment in
+  Step 5.
+
+Parse each sub-agent's JSON and keep only the compact result. If a sub-agent's output
```
jwbron, line 227: The orchestrator no longer imports `risk-classification.md`, `ci-tooling.md`, or `skills.md`. Those exist only inside the sub-agent sections, So the orchestrator lacks the criteria to redo risk or skills analysis. Either drop the promise (degrade to "skip that dimension and note it in the review") or also import the configs into the orchestrator. As written it's a false safety net.

jwbron (changes requested):
A definite improvement but there are some tweaks I would make here, mainly consideration about what models are being used. I would strongly suggest bumping some of these reviewers up to opus. With sonnet, we're getting lower quality reviews and being given what could come across as a false sense of security. I understand the cost concerns here, but I'm not sure the solution of having critical checks be done by models that don't have sufficient reasoning capabilities is the right move.

I would also suggest adding some generalized verbiage like "every line in the diff you are given must be read, this review must be comprehensive" that's included in all reviewer prompts (excluding targeted reviewers that use deterministic checks to find relevant changes).

question: disk-staging depends on tooling that isn't explicitly declared. Does the claude engine grant Write/Bash by default, or does this need declaring? This could be verified via compilation.

jeresig:
@claude review once

jeresig:
@jwbron Thanks for the feedback! I believe I've addressed your comments (raising the model quality in almost all cases, providing clearer instructions to read every line).

And for disk access, it does have access to the various tools that were granted to the parent agent (so Read/Bash are included). I can verify this in the frontend repo before landing though!

jeresig:
@claude review once

jwbron (approved):
LGTM after addressing claude's feedback. Some more feedback that doesn't need to be addressed here but should probably be a followup:

Consider having a separate agent validate all claims. This helps reduce false positives. Probably only needed if we see regular noise caused by incorrect or misleading claims.

It seems this always operates on the full PR diff, not the diff since the last review. This will cause some noise and increased costs due to having the same code continually re-reviewed. I'd suggest updating this to be aware of what's already been reviewed, only presenting diffs with new/updated code.

This is overall a diff-centric workflow, so broader context can be missed. I have a "holistic reviewer" that steps back and looks for issues that could be caused in existing code, e.g. looking at existing callers for a function that was updated. Could be worth adding another agent for this.

Not sure about this one, but it could be worth dropping the blocking posture until this is a bit more dialed. Your call here though since this is already being used in frontend.

somewhatabstract (approved):
This generally looks good to me, based off a general pass and relying on @jwbron review and Claude's comments.

jeresig:
@jwbron I added in a claim-validator sub-agent that attempts to validate all comments before they are left. Hopefully that helps a bit!

PR #195, opened 2026-07-02
apply-terraform-plan: clean up plan files via PR, not a direct push

csilvers:
## Summary:
The "Delete Plan Files" step committed the plan-file cleanup and pushed it
straight to master. Now that master is a protected branch requiring PRs,
that push is rejected (GH006) and the apply job fails even though the
terraform apply itself succeeded.

Stage the deletion and open a PR instead, mirroring how
generate-terraform-plan lands its plan updates. Base defaults to the branch
the workflow ran on, so this also works for repos whose base branch is main.

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

Issue: INFRA-10923

## Test plan:


jeresig (approved):
Seems reasonable to me!

jwbron:
I was reviewing this along with claude code and for some reason it was extraordinarily comprehensive. I've shortened it a bit and removed some irrelevant stuff, but here's its feedback. Looks solid to me, see the last line for a summary.

The direction is right, but the flow stops one step short of automation, and that missing step is the root of most findings. Relevant facts it verified: internal-webserver's master protection requires PRs but 0 approving reviews and no required status checks, and auto-merge is disabled on the repo.

1. Cleanup is no longer immediate: the already-applied plan files sit on master until a human merges the cleanup PR. In that window, a job re-run or any push matching the apply workflow's path filters re-runs terraform apply on the consumed plan, fails with "saved plan is stale", and posts a failure comment on an unrelated PR.
2. A human merging the cleanup PR pushes a commit touching tfplan.binary, which is in the consumer's `on.push.paths`, so every apply cycle spawns a second apply run (the old GITHUB_TOKEN push triggered nothing). That run skips everything but concludes SUCCESS, so generate-terraform-plan's "Changes Since Last Deployment" compare link points at the cleanup merge instead of the actually-deployed commit.
3. No cleanup-PR lifecycle management: each run creates a unique `ci/terraform-cleanup-<run_id>` branch and nothing closes superseded cleanup PRs, so once a newer plan PR merges, older unmerged ones become permanent modify/delete conflicts. A fixed branch name would let create-pull-request update a single PR in place.
4. The design adds a human gate but doesn't wire up the human: no `reviewers:` is set, so nobody is notified the PR exists. Better: skip the human entirely. With 0 required approvals and no required checks, a follow-up step can `gh pr merge --squash` the PR immediately with the same GITHUB_TOKEN (auto-merge is disabled, so `--auto` won't work; merge directly, with a short retry while GitHub computes mergeability). A GITHUB_TOKEN merge also doesn't trigger push workflows, which eliminates findings 1-3 in one stroke.
5. The cleanup steps are gated only on `cleanup_plan_files`/`auto_approve`, not on `steps.apply.outcome == 'success'`, while the apply step has `continue-on-error: true`. A failed apply still opens a PR whose body says the plan "was applied"; merging it discards a reviewed-but-never-applied plan.
6. No `base:` is passed and the action exposes no input for it; the pinned create-pull-request errors on detached-HEAD checkouts. Only bites future call sites (current consumers use push-event checkouts), but the sibling action passes `base:` explicitly; mirror that.
7. Polish: missing `delete-branch: true` leaves one dead `ci/terraform-cleanup-*` branch per apply, and the `git rm -f` step is redundant with `add-paths` (which stages deletions itself; a plain `rm -f` is also more robust when one file is missing).

Suggested shape of the fix: keep the PR-based flow, add `gh pr merge --squash` (with retry) right after creation, gate cleanup on `steps.apply.outcome == 'success'`, use a fixed branch name plus `delete-branch: true`, and pass an explicit `base`.

csilvers:
OK, I'm just putting all that into claude and telling it to fix things.  We're out of my depth I'm afraid.

csilvers:
OK, I think the PR I (well, claude) just pushed has the fix as suggested!

jwbron (approved):

PR #205, opened 2026-07-03
[pipeline-dcdad92d][slice-10/12] Wave-2 recall/precision rebalance

jwbron:
Land edits 8-13 and the posting bar, verified against the smoke set so recall/precision do not regress while the rebalance ships. Includes the webapp #40536 causal experiment via the no-post harness.

**Revised 2026-07-07:** the blocking-claim refuter panel originally in this slice is **removed rather than shipped**, per the review discussion (https://github.com/Khan/actions/pull/205#discussion_r3538111943) and a production audit of every blocking review the bot posted on Khan/frontend since the validator landed (90 PRs, 319 blocking claims, 2026-06-30 to 2026-07-07): 12 confirmed false blocks, of which only 2 were in the panel's addressable class (the rest were a repeated mechanically-refutable claim, severity misreads, and author-intent questions no refuter could adjudicate from code). In its place, at zero extra agents:

- **Three-state claim validation** (`confirmed` / `plausible` / `refuted`) with symmetric evidence duties: confirm by citing the failing line, refute by citing the guard. **Only a confirmed claim may carry a blocking label into the computed verdict**; plausible downgrades to non-blocking.
- **Author-dispute usage-depth bar**: a claim the author has factually disputed cannot re-block unless the re-check traces to actual usage (the frontend #13214 failure shape, where two checks each stopped one parent short); otherwise it posts as a question.
- **False-block tripwire in the eval suite**: the no-post runner replays the gate's apply rules deterministically (stage 3b), and the smoke corpus gains six audit-seeded cases: five production false blocks that must APPROVE, one confirmed block that must stay REQUEST_CHANGES. If the false-block metric ever regresses, the refuter is recoverable from this PR's history.

**Base PR:** #196

## What's in this PR

```text
 .changeset/review-roster-framework.md                        |   2 +-
 .changeset/review-wave2-rebalance.md                         |   2 +-
 workflows/review/eval/corpus/loader.ts                       | 105 ++++++++
 workflows/review/eval/corpus/smoke/confirmed-block-survives-gate.json | 36 ++
 workflows/review/eval/corpus/smoke/downgrade-author-intent.json       | 35 ++
 workflows/review/eval/corpus/smoke/false-block-api-contract.json      | 36 ++
 workflows/review/eval/corpus/smoke/false-block-author-disputed-h1.json| 36 ++
 workflows/review/eval/corpus/smoke/false-block-contrast-token.json    | 36 ++
 workflows/review/eval/corpus/smoke/false-block-convention-misread.json| 36 ++
 workflows/review/eval/runner.ts                              |  73 +++++-
 workflows/review/eval/smoke.test.ts                          | 241 ++++++++++++
 workflows/review/review.md                                   | 268 ++++++++++-----
 12 files changed, 833 insertions(+), 76 deletions(-)
```

## This slice

Wave-2 recall/precision rebalance

Files affected:
- `workflows/review/review.md`
- `workflows/review/eval/smoke.test.ts`
- `workflows/review/eval/runner.ts`, `workflows/review/eval/corpus/` (validation replay + audit-seeded cases)

<details>
<summary>Tasks (4) + acceptance criteria</summary>

- task-10-1: Edits 8 (coverage first), 9 (blocking requires a concrete failing scenario), 10 (drop only the refuted; downgrade the uncertain), 11 (confirm before you claim), 12 (cite exact lines or quote).
  - Acceptance criteria: Each edit present.
- task-10-2 (revised): Edit 13 posting bar (ranked posting; inline >= medium confidence; low-confidence in one collapsed section; suggested diffs where clear; no padding) wired to the slice-2 verdict / slice-1 confidence fields. The blocking-claim refuter panel is replaced by the three-state validation gate (only confirmed claims block), per review + the production audit above.
  - Acceptance criteria: Posting bar + confirmed-only gate wired to verdict/confidence.
- task-10-3: Verify the rebalance against the slice-9 smoke set (no recall regression on must-catch; no new false-block on clean).
  - Acceptance criteria: Smoke set green after rebalance.
- task-10-4: R6 causal experiment: rerun the reviewer on webapp PR #40536 with edits 8+10 via the no-post harness (no consumer write). Success = the OpenAccess authorization question surfaces.
  - Acceptance criteria: Experiment recorded; surfacing observed or a documented negative result.

</details>

## Stack

- Position: slice 10 of 12 in pipeline `pipeline-dcdad92d`
- Base PR: #196
- Stacked on top of `egg/pipeline-dcdad92d/slice-9`

## Causal experiment: webapp PR #40536 (OpenAccess authorization)

Formerly `eval/experiments/webapp-40536.md`; the file is removed from the repo per review feedback and the record lives here instead.

**Hypothesis:** the prior reviewer missed the authorization question on Khan/webapp #40536 (a resolver made `OpenAccess`, its guard effectively removed) as a recall failure that this slice's edits fix: E8 (coverage first) makes the security-auth reviewer raise the concern instead of self-censoring; E10 (drop only the refuted; downgrade the uncertain) stops the claim-validator from silently dropping it.

**Method (no consumer write):** the slice-9 no-post runner replays the real deterministic path (`router.route` -> `labelForFinding` -> scope filter -> `computeVerdict` -> render) and returns the review it would submit as plain data. The #40536 OpenAccess pattern is encoded as the smoke corpus case `incident-auth-bypass` (`src/auth/middleware.ts:57`, the admin guard's early `return 403` removed), run with recorded findings: deterministic, no model, no network.

**Result:** with edits 8+10 applied, the authorization question surfaces as a posted blocking finding on the stand-in case; the pre-rebalance stance dropped it. The live rerun against the actual webapp PR is a documented follow-up owned by the consumer rollout (`runCase(corpusCase, {produceFindings})` swaps in a live producer with every downstream stage identical).


workflows/review/review.md
```diff
@@ -643,6 +672,42 @@ output is missing or unparseable, do **not** drop the comments: post the unvalid
 claims anyway, and surface the gap as a skipped dimension (`claim validation`) with the
 note in Step 6, so the author knows they were not double-checked this run.
 
+**Phase 3b — refute the blocking claims (blocking-claim refuter panel).** After the
```
jeresig: I'm really not sure we need this - imo, if we determine that the claim validator isn't enough then we can escalate (but I feel that it's already the case that it's fine).

workflows/review/eval/experiments/webapp-40536.md
```diff
@@ -0,0 +1,92 @@
+# Causal experiment — webapp PR #40536 (OpenAccess authorization)
```
jeresig: I'm not sure what this and the test are in reference to?

jeresig (changes requested):
I think that this change isn't really necessary - we already have a validator in place - I think that's more than sufficient, for now.

workflows/review/review.md
jwbron: I'd push back on "the validator is already enough". The frontend production record in the proposal doc has two cases where a wrong claim survived validation and blocked anyway (the refuted exhaustive-deps claim on #12249, and the a11y claim re-raised across pushes after the author disproved it on #13214). So the escalation trigger you're describing has already fired. And this slice makes it more likely, not less: the recall edits deliberately soften the validator (downgrade the uncertain instead of dropping it), so the load on the blocking path goes up exactly where a wrong call costs the most trust (a wrong REQUEST_CHANGES, 44% of frontend's blocks were the over-block class).

Cost-wise this phase is close to free. It runs only over claims that still carry a blocking label after validation, which is typically 0-3 per PR, and it's skipped entirely when there are none. The reason it's a separate agent rather than more validator prompt is independence. The validator re-grading a claim it just decided to keep is grading its own homework, while the refuter's only mandate is to break the claim.

That said, if you still disagree, we could meet in the middle. I could shrink it from a panel to a single adversarial refuter pass over the blocking survivors and add a false-block metric to the eval suite so this stage has to earn its line like everything else. If the numbers say the validator alone holds, removing it is a one-paragraph deletion.

jeresig: Just to clarify though, we only just landed the new one with the validator late on Thursday - all reviews that went before that landed (and some after) didn't have any validation step. I don't think it's really been run much at all. #12249 was from last month and #13214 was also before the validator went out. I don't think we have enough data at all to know if this is needed.

jwbron: Fair point, I'll go ahead and drop the panel (it's in this PR's history if we ever need it back). This PR stays a rebalance of the finders and the claim validator: recall-tuned finders, three-state validation (only CONFIRMED claims can block, uncertain ones downgrade to non-blocking), and a false-block metric in the eval suite as the tripwire. I audited all ~320 blocking claims since the validator shipped and the false blocks cluster in repeat claims and severity misreads, not things another verification layer would catch, so agreed it isn't earning its cost. If the metric ever moves we can escalate then, with data.

jeresig (approved):
Great, thanks for reworking this.

PR #210, opened 2026-07-07
fix-workflows: support inline `fix-workflows-ignore` comments

jaredly:
## Summary

The `fix-workflows` action rewrites every `runs-on:` to the conditional `ephemeral-runner` expression (when `fix-runs-on` is enabled) and inserts a secure-network setup step after every `actions/checkout`. Some jobs legitimately need a specific runner — e.g. [Khan/webapp#40211](https://github.com/Khan/webapp/pull/40211) needs to route go-test shards to a dedicated runner and, more generally, a job may need `runs-on: ubuntu-latest-m` — or intentionally skip the setup step, but there was previously no way to opt out.

This adds an inline escape hatch, anchored to the exact line being changed (like `# noqa` / `// eslint-disable-line`):

- **runs-on:** `runs-on: ubuntu-latest-m # fix-workflows-ignore` — skips the runs-on rewrite for that job.
- **setup step:** `uses: actions/checkout@v4 # fix-workflows-ignore` — skips inserting the setup step after that checkout.

Directives can be **unscoped** (skip every rule for the annotated line) or **scoped** to a specific rule: `# fix-workflows-ignore: runs-on`, `# fix-workflows-ignore: setup`, or `# fix-workflows-ignore: runs-on, setup`. `# lintignore` is accepted as an alias. A directive may also be placed as the leading comment of the job (for runs-on) or the checkout step (for setup) instead of trailing the line.

## Changes
- **`index.ts`** — parse `fix-workflows-ignore` directives from YAML comments (`comment` / `commentBefore`) and honor them: `checkRunsOn` (which also gates `fixRunsOn` and the verify pass) returns "compliant" when the runs-on line opts out; `fixSteps`/`checkSteps` skip any checkout whose step or `uses:` line opts out. New exported helpers `runsOnIgnored` and `stepIgnoresSetup`.
- **`index.test.ts`** — coverage for trailing and leading placements, scoped vs unscoped directives, the `lintignore` alias, per-checkout granularity (one ignored checkout doesn't exempt a sibling), and the `processJob`/`checkJob` integration paths.
- **`action.yml`** — mention the opt-out in the action description and in the per-file "needs updating" warning.
- **`.changeset/`** — minor changeset (the release workflow bumps the version and CHANGELOG on landing).

## Verification
- `npx vitest run` — 183 passing (55 in `fix-workflows`, 17 new).
- `tsc --noEmit` and `eslint` clean.
- End-to-end run of the action's entrypoint against a fixture: a plain `runs-on: ubuntu-latest` job is fully rewritten while a `runs-on: ubuntu-latest-m # fix-workflows-ignore` job (with an ignored checkout) is left untouched, comments preserved, and the post-fix verify pass reports no remaining violations.

https://claude.ai/code/session_01TP92sbdHXuVE5skDCAt2fm

---
_Generated by [Claude Code](https://claude.ai/code/session_01TP92sbdHXuVE5skDCAt2fm)_

### DIFF
--- a/actions/fix-workflows/index.ts
+++ b/actions/fix-workflows/index.ts
@@ -10,6 +10,9 @@
  * When a violation is found it is fixed automatically.
  * Comments are preserved via the yaml package's document API.
  * Run oxfmt after this script to normalize formatting.
+ *
+ * Individual jobs, `runs-on:` lines, or checkout steps can opt out of a fix
+ * with an inline `fix-workflows-ignore` comment — see IGNORE_DIRECTIVE_RE.
  */
 import * as fs from "node:fs";
 import * as path from "node:path";
@@ -19,6 +22,7 @@ import {
     isSeq,
     parseDocument,
     type Document,
+    type Node,
     type YAMLMap,
     type YAMLSeq,
 } from "yaml";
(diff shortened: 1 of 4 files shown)


jaredly:
@jeresig what do you think of this approach? If we want to remove the `ignore setup` part I'm fine with that -- runner flexibility is the primary aim

actions/fix-workflows/index.ts
```diff
@@ -33,6 +37,143 @@ const YAML_WRITE_OPTIONS = {indent: 4, lineWidth: 0} as const;
...
+function normalizeRule(token: string): IgnoreRule | null {
+    switch (token.trim().toLowerCase()) {
+        case "runs-on":
+        case "runson":
+        case "runs_on":
+            return "runs-on";
+        case "setup":
+        case "secure-network":
+        case "checkout":
+            return "setup";
```
jeresig, line 73: oof yeah, I'm hesitant about this one. I think there could, theoretically, be cases where we don't care about this (e.g. stuff running when it's on main where it's already gone through code review) but it's hard to get right. I think we can keep this and just not advertise it, lol

jeresig (approved):
Great! I think I saw a case in frontend where we had to do something similar to force the runs-on. Nice to have an official solution!

actions/fix-workflows/index.ts
jaredly: hah fair

PR #207, opened 2026-07-03
[pipeline-dcdad92d][slice-12/12] P2 items (R13, R14, R15, R16, R17)

jwbron:
Land the P2 follow-ups: per-finding resolution rule, the drift-guard documentation reusing the version stamp, live counters, dismissal-learning candidates, and the conditional-approval verdict.

**Base PR:** #196

## What's in this PR

Commits (5):
- Persist BRC history for slice-12 (#2548)
- implement(coder): slice-12 — fix NUL-byte encoding in dismissal-learning groupKey
- implement(coder): slice-12 P2 — live counters, dismissal-learning, conditional-approval
- implement(documenter): slice-12 R14 — correct drift-guard doc (reviewer_code NACK v1)
- implement(documenter): slice-12 — R13 per-finding re-review resolution + R14 drift-guard doc

```text
.../brc-history/pipeline-dcdad92d-implement-slice-12.json       | 2133 +++++++++++++++++++++++
 .egg-state/brc-history/pipeline-dcdad92d-implement-slice-12.md  | 2433 +++++++++++++++++++++++++++
 workflows/review/README.md                                      |   36 +
 workflows/review/lib/counters.ts                                |  486 ++++++
 workflows/review/lib/dismissal-learning.ts                      |  357 ++++
 workflows/review/lib/render-comment.ts                          |  123 +-
 workflows/review/review.md                                      |   13 +
 7 files changed, 5576 insertions(+), 5 deletions(-)
```

## This slice

P2 items (R13, R14, R15, R16, R17)

Files affected:
- `workflows/review/review.md`
- `workflows/review/README.md`
- `workflows/review/lib/counters.ts`
- `workflows/review/lib/dismissal-learning.ts`
- `workflows/review/lib/render-comment.ts`

<details>
<summary>Tasks (5) + acceptance criteria</summary>

- task-12-1: R13 per-finding resolution rule on re-review: every actionable finding gets fixed / deferred-to-filed-issue / disagreed-with-reason.
  - Acceptance criteria: Rule present.
- task-12-2: R14 config drift guard: document the slice-11 version stamp as the stable consumer-readable sync surface (interface §4.5). Adds NO new mechanism.
  - Acceptance criteria: Doc points to the stamp; no second surface.
- task-12-3: R15 live counters mined from run logs + #194 per-run JSON artifacts: validator drop rate per lens, comments/PR, verdict mix, thumbs agree rate, cost/run.
[...]

PR #212, opened 2026-07-07
review: read per-directory REVIEW.md contracts from the consuming repo

jwbron:
Stacked on #207.

## Problem

webapp documents sub-tree-specific review expectations in `REVIEW.md` files (a root one plus one per documented top-level directory, next to each `AGENTS.md`), but nothing in the review workflow ever reads them. They are not pulled in by default:

- gh-aw runs the Claude Code CLI headless (`claude --print`, no `--bare`), so the repo's `CLAUDE.md` (webapp symlinks it to `AGENTS.md`) IS loaded for the orchestrator and inherited by the inline sub-agents.
- But `AGENTS.md` references the contracts as plain Markdown links, not `@`-imports, so they are never inlined; `REVIEW.md` is not a memory filename to Claude Code; and the sub-agent prompts point only at the diff, the `.github/aw/review/` runtime imports, and the skill files.

So the review contract only reached the reviewer if a model happened to follow a link.

## Change

- `correctness-reviewer`: when the checkout carries `REVIEW.md` files, read the root contract plus the nearest `REVIEW.md` above each reviewed file, and use them to sharpen `riskReason` wording and calibrate finding severity for that sub-tree.
- `claim-validator`: use the same contracts to calibrate claim labels (a contract calling a category a nit supports correcting an overstated label). Contracts never decide `verification`; that stays code-evidence-only.
- Both steps are conditional on the files existing, so frontend and other consumers without `REVIEW.md` files are unaffected.
- README: document the convention and its trust boundary.

## Trust boundary

Unlike `.github/` and the agent config folders (restored from the base branch by gh-aw before the agent runs), `REVIEW.md` files are read from the PR head. The prompts therefore treat contract text as guidance that can adjust emphasis but never override the workflow's rules, and an edit to a `REVIEW.md` in the diff is reviewed on its merits under the existing steering-text rule.


jeresig (approved):
Seems reasonable!

PR #215, opened 2026-07-08
review: unblock first-principles (Fable 5 pricing) and harden the artifact-upload path

jwbron:
Fixes two failures observed on the first production run of review-v1.2.0 (webapp, Khan/webapp#40678):

1. **first-principles failed on every dispatch.** Its pinned model `claude-fable-5` is not in the AI-credits pricing table bundled with gh-aw <= v0.81.x, so the cost-guardrail API proxy rejected the request with `400 Model "claude-fable-5" has no AI credits pricing` before it reached the model (11 instant retries, 0 tokens). The frontmatter now merges the model's pricing via gh-aw's `models:` field, using the same numbers upstream added to gh-aw-firewall v0.27.27 ($10/M input, $1/M cache read, $12.50/M cache write, $50/M output; expressed per token as the field requires). The block is commented for removal once the workflow runs on a gh-aw release whose bundled table includes the Claude 5 family.

2. **The Step 9 artifact upload failed validation.** The orchestrator ran the upload from inside `/tmp/gh-aw/review/out` and recorded the relative path `out`, which the safe-outputs processor could not match against `allowed-paths` (`/tmp/gh-aw/review/out/**`), producing `ERR_VALIDATION: no files matched the selection criteria` (the files still reached the artifact via staging, so this was an error annotation rather than data loss). The instruction now requires the absolute path explicitly.

Also bumps the lib-checkout ref to `review-v1.2.1`, the release this ships in.

### DIFF
--- a/.changeset/review-fable-pricing-upload-path.md
+++ b/.changeset/review-fable-pricing-upload-path.md
@@ -1,0 +1,5 @@
+---
+"review": patch
+---
+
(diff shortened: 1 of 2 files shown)


jeresig (approved):
Great, thank you!

PR #217, opened 2026-07-08
review: pin the firewall to v0.27.27 so the api-proxy prices claude-fable-5

jwbron:
Follow-up to #215, which did not actually fix the first-principles failure: the `models:` frontmatter block feeds gh-aw's cost-summary display (`GH_AW_INFO_MODEL_COSTS`) but never reaches the firewall api-proxy's AI-credits guard, so the proxy kept rejecting `claude-fable-5` with a 400 (verified on the next run of Khan/webapp#40678).

The real fix is `sandbox.agent.version: v0.27.27`, which runs the gh-aw-firewall release whose api-proxy bundles curated Claude 5 pricing (the same numbers #215 added for display). Verified by compiling against gh-aw v0.81.6: the agent job's awf-config `imageTag` moves from 0.27.11 to 0.27.27.

The `models:` block stays for cost-display consistency; both pieces are commented for removal once gh-aw's default firewall is >= v0.27.27. Lib-checkout ref bumped to `review-v1.2.2`, the release this ships in.

Not addressed here: the `upload_artifact` ERR_VALIDATION annotation is a gh-aw bug (the safeoutputs MCP server records the staging-relative path `out` even when the agent passes the absolute path, and the processor then validates that recorded path against `allowed-paths`); it is harmless (the staging step still uploads the files) and needs an upstream github/gh-aw issue rather than a prompt change.

### DIFF
--- a/.changeset/review-firewall-pin-fable.md
+++ b/.changeset/review-firewall-pin-fable.md
@@ -1,0 +1,5 @@
+---
+"review": patch
+---
+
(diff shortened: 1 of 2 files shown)


jeresig (approved):
Tricky!

PR #225, opened 2026-07-09
review: turn on the review-feedback signal (thumbs-sweep GitHub port + live-counters report)

jwbron:
Was stacked on #223 (thumbs-sweep seeding + self-reaction filtering); #223 has landed and main is merged in.

Implements the Khan/actions side of the design doc's "The tuning loop" -> "Signal to turn on": the production GitHub wiring for the thumbs sweep and the live counters. Consumer-repo PRs (webapp/frontend scheduled workflows, frontend ROUTING) follow separately and pin the release this ships in.

## Thumbs sweep

- **`lib/thumbs-sweep-github.ts`** — the octokit-backed `ThumbsSweepPort`. Reviewer comments are identified per grain by the workflow's markers, not by author alone (`github-actions[bot]` authors many workflows' comments):
  - *summary*: issue comments carrying the `pr-reviewer:risks-and-patterns` marker **or** gh-aw's engine-emitted `gh-aw-workflow-call-id: <owner>/<repo>/review` marker. Verified against live comments: production summary comments carry only the latter today.
  - *inline*: review comments whose body starts with one of the code-owned Conventional-Comment label prefixes (built from `render-comment.ts`'s taxonomy). Inline comments carry no hidden marker at current releases; the label grammar is code-owned and templated, so the match is exact.
  - Reactions are listed per comment **with reactor logins**, so the self-reaction filtering excludes the bot's own reactions (the post-time seeded nudges, once those exist) from all signal. Live tallies use the shared positive/negative sets (👍/❤️/🎉/🚀 vs 👎/😕, matching gh-aw's outcome evaluation).
  - **Resolved inline threads are counted** (one GraphQL `reviewThreads` query per PR that has reviewer inline comments) and reported as their own positive column; resolution never triggers a follow-up.
  - Bounded: PRs updated in the last 14 days (configurable), capped at 200, **skipping PRs closed/merged more than 3 days ago** (`REVIEW_SWEEP_CLOSED_GRACE_DAYS`; feedback lands around merge time, after which a landed PR stops changing). Reaction detail is fetched only when a comment has any reactions. Every API request is counted and reported.
- **`lib/run-thumbs-sweep.ts`** — env-configured CLI for the consumer workflows (`GITHUB_TOKEN` + `pull-requests: write` is all it needs). Writes the `SweepResult` JSON to stdout and an auditable Markdown digest to the job summary. `REVIEW_SWEEP_DRY_RUN=true` traverses and decides without posting, for first-run audits.
- The `review` package gains its first pinned runtime dependency, `octokit` (throttling/retry built in), with the transitive tree locked by a committed `package-lock.json`. Consumer sweep workflows run `npm ci --omit=dev` in the checked-out `workflows/review/`; all other lib scripts remain dependency-free. Loaded via dynamic `import()` because the package is ESM-only and tsx runs the lib as CJS.

## Live counters

- **`lib/counters-report.ts`** — aggregates a directory of downloaded per-run artifacts into the weekly counters job summary via `counters.ts`. Where `summary.json` doesn't exist (all current production runs), it synthesizes the run summary from gh-aw engine artifacts every run already uploads: `safeoutputs.jsonl` (verdict event), `safe-output-items.jsonl` (comments actually posted), `agent_usage.json` (cost/tokens). Artifact files are located by basename at any depth, tolerating the known gh-aw staging-path nesting.
- **`lib/counters.ts`** — normalization extended to the artifact shapes production actually emits: validator files keyed `{"claims": [...]}` with three-state `verification` (`confirmed`/`plausible` fold to keep, `refuted` to drop), and an opt-in fallback source (`(unknown)`) for decisions that can't be joined to `claims.json` (which production runs don't upload). Default join behavior is unchanged.

## Changes from review feedback

- **Sweep-time reaction seeding is removed** (the core's `seedReactions`, the port's `addReactions`, `REVIEW_SWEEP_SEED_REACTIONS`). The nudge pair will instead be seeded at post time by a custom safe-output job in the consumer repos' review workflow (follow-up PR); the sweep's self-reaction filtering stays, keeping those seeds out of all signal.
- **Resolved threads count as success**: counted per PR via GraphQL, reported as a separate positive column (threads also get resolved just to clear noise, so resolutions aren't folded into 👍).
- **Wider reaction sets in the live tally**: the reported numbers now use the same 👍/❤️/🎉/🚀 vs 👎/😕 sets the follow-up logic already acted on.
- **Closed/merged PRs age out**: swept for a 3-day grace window after closing (captures the around-merge feedback tail), then skipped; keeps the sweep stateless.
- **README section** positioning the sweep vs the gh-aw outcome-collector (passive Sentry telemetry vs active reason elicitation), including the documented `add_comment` inflation once post-time seeding exists.

## Verification

- 750 lib tests pass (post-merge with main; includes a new `renderSweepSummary` unit test), typecheck and lint clean.
- Live dry-run against both consumer repos (3-day window, 15 PRs each): Khan/frontend identified 72 inline + 24 summary reviewer comments, Khan/webapp 14 + 2, 31 API requests each, zero false matches spot-checked; write paths exercised end-to-end in tests against a fake GitHub. (Dry-run predates the resolved-thread/grace-window changes; the traversal shape is unchanged.)

No review semantics change anywhere; the reviewer's prompt and workflow are untouched.

cc @jeresig

## Human steps (non-review)

- [ ] **Schedule it in the consuming repo**: webapp#40711 (deploy-branch flow) carries the actual schedule; land it via the normal deploy process after this merges.
- [ ] **Watch the first scheduled sweep and weekly counters report** end to end; note that artifact-reading counters depend on the `out/` upload fix (#243) reaching a release before they see claim data.


jeresig (changes requested):
I think this is running counter to what we already have going with the gh-outcome-collector logic. I think we want to be using that instead of this. Or, this needs *much* better documentation as right now it's really not clear how this is being run and when and how the data is being used (or not). Like, is this being integrated into our existing OTel data?

If we are using this we need to integrate the other ways that we're determining success (such as a resolved thread = success, and also other reactions equaling success/rejection). I raised this in your other PRs already.

jeresig (changes requested):
Ok, now that I see how this is being used, a number of changes requested:

* Resolved threads should be counted as success.
* Closed or merged PRs should only be analyzed once (they aren't going to be changing once they've landed)
* We should take into account other affirmative/negative reactions (I raised this in another PR)

Question on leaving the default thumbs-up/down on the review comment - are you doing that here as it's not possible to do in the gh-aw safe outputs?

jwbron:
Thanks, responding to both reviews together.

Outcome-collector overlap: agreed it needs explicit positioning. The sweep isn't replaceable by the collector: the collector never writes to GitHub, so it can't ask why a comment was downvoted, and those reason labels (incorrect / unimportant / unclear / duplicate) are what calibrate the eval judge. I'll add a README section spelling out the division: collector = passive acceptance telemetry to Sentry; sweep = active reason elicitation, results in the job summary + stdout JSON (not otel today; I can export the tallies over the same OTLP endpoint review.md uses if you want that).

Resolved threads = success: agreed. Matches the gh-aw spec; notably the JS evaluator we run doesn't implement it yet either (only the Go CLI has the dedicated evaluator). I'll add a GraphQL reviewThreads pass and report resolutions as their own positive column rather than folding them into 👍, since threads also get resolved just to clear noise.

Closed/merged analyzed once: partly agreed. The sweep is deliberately stateless, and feedback often lands right around merge, so I'd rather skip PRs closed more than ~3 days ago than track analyzed-once state. Same effect at current scale (~31 API requests/run) without a state store.

Other reactions: agreed, half done. The core sets already match the collector's (👍 ❤️ 🎉 🚀 / 👎 😕) and the follow-up already triggers on 😕; the live tally only counts literal +1/-1, which I'll fix to use the shared sets.

Seeding question: close. gh-aw can't do it natively (the comment IDs don't exist until the handlers post them, and GITHUB_TOKEN-posted comments don't emit events, so an event-triggered seeder never fires). But a custom safe-output job (safe-outputs.jobs with needs: safe_outputs) can react seconds after posting by reading the handler's safe-output-items record. So I'll drop REVIEW_SWEEP_SEED_REACTIONS from this PR and do post-time seeding in the consumer review.md instead.

One interaction to flag: seeded pairs make the collector read every review summary comment as accepted (its add_comment evaluator counts any reaction, identity-blind). The damage is bounded to that one metric (inline comments use an existence fallback in JS; submit_pull_request_review is scored by commits/merge state), and the sweep measures reviewer-comment engagement strictly better (identity-filtered, both grains, reason labels). Plan: seed anyway, document that the collector's add_comment result for the review workflow is inflated by design with the sweep authoritative there, and file an upstream gh-aw issue for identity-aware reaction counting (their spec already prefers bot-identity provenance). Shout if you'd rather hold seeding until upstream fixes it.

So for this PR: drop sweep-time seeding, count resolved threads, widen the reaction tally, closed-PR grace window, README section on the two systems. Follow-ups: post-time seeding in the consumer repos, upstream gh-aw issue. Sound right?

jwbron:
Changes from the review are in (7c601d3): sweep-time seeding removed (post-time seeding follows in the consumer repos), resolved inline threads counted via GraphQL as their own positive column, the live tally widened to the shared reaction sets (👍/❤️/🎉/🚀 vs 👎/😕), closed/merged PRs age out after a 3-day grace window, and the README now has a section positioning the sweep against the outcome-collector (including the add_comment inflation caveat). PR description updated to match. 298 lib tests pass, typecheck and lint clean.

jeresig (approved):
Great - thanks for the fixes!

jwbron:
Merged main (pnpm-lock regenerated; `octokit` + main's `claude-agent-sdk` coexist) and addressed the bot's non-blocking feedback in ab0b556:

- **Lockfile**: `workflows/review/package-lock.json` is now committed (generated in isolation, registry-resolved), so the exact `octokit` pin also locks the transitive tree. Docstring and README updated to `npm ci --omit=dev`; the consumer workflows (webapp#40711) will use `npm ci` too.
- **`renderSweepSummary` test**: added `run-thumbs-sweep.test.ts` covering the dry-run vs normal header, the already-followed-up count, and the conditional downvoted table; the "tests import `renderSweepSummary`" comment is now true.
- **Verdict precedence**: the synthesized verdict now prefers REQUEST_CHANGES if a malformed artifact carries multiple review submissions, with a test.

Skipped two:
- The `closed_at` fallthrough is defensive-only (GitHub always populates it for closed PRs); worst case is a few extra reads.
- The repo-wide comment-listing traversal is a real efficiency idea but a redesign of the traversal shape post-approval; noting it as a follow-up candidate if sweep API usage ever matters (today's sweeps run ~31 requests).

750 lib tests pass, typecheck and lint clean.

PR #226, opened 2026-07-09
review: require a concrete failure_scenario on every finding

jwbron:
## Summary

First of three PRs split out of #224 (stack: this PR, then #227 provenance gate, then #228 prompt disciplines). Each lands independently revertable; the three together are byte-identical to #224.

Every finding (not just blocking ones) now carries a concrete `failure_scenario`: the specific inputs or state and the wrong outcome they produce. The field is required in the structured finding schema (`FINDING_SCHEMA_VERSION` bumped to 2), emitted by every producer (label-shape reviewers and all 11 specialist lenses), and carried verbatim into `claims.json`. The claim-validator now attacks exactly that stated scenario: confirmed means the named scenario occurs on the cited lines, and a scenario too vague to check caps at plausible.

`FINDING_SCHEMA_VERSION` is now 2 (a required field invalidates previously serialized findings, per the schema module's own rule); every corpus fixture is migrated with a hand-written failure scenario.

cc @jeresig

## Testing

- `npx vitest run workflows/review`: 410 tests green, including the smoke set.
- `npx tsc --noEmit` clean.

### DIFF
--- a/.changeset/review-failure-scenario.md
+++ b/.changeset/review-failure-scenario.md
@@ -1,0 +1,5 @@
+---
+"review": minor
+---
+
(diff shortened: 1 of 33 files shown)


jeresig (approved):
Great!

PR #227, opened 2026-07-09
review: enforce the change-provenance gate in code; stage generated-stripped diffs

jwbron:
## Summary

Second of three PRs split out of #224; stacked on #226 (retarget to main when it merges). The follow-up prompt-disciplines PR stacks on this one.

**Change-provenance gate, enforced in code.** New `lib/diff.ts` (unified-diff parsing) and `lib/provenance.ts` (gate + CLI) compute a per-file changed-line map (`provenance.json`). A finding whose anchor is not an added or modified line of the diff cannot carry a blocking label, and pre-existing observations collapse into at most one non-blocking collapsed note (`renderPreExistingNote`). Deletion findings pass via `removedAdjacent` lines. The gate fails open (gates nothing, with a review-body note) whenever the parsed map cannot be trusted: an unparseable diff, hunk headers not attributable to any file section, or a changed file whose patch is missing from the parse (`files.json` now carries a `hasPatch` flag so binary/too-large files do not false-positive the check). Wired through the no-post eval runner and covered by unit tests plus a new smoke corpus case (`provenance-pre-existing-note`).

**Generated-stripped whole-change diffs (measured cost fix).** The provenance CLI also stages `full-stripped.diff` (the full diff minus files the router classifies `linguist-generated`; `routing.json` now exposes `generatedFiles`). Every whole-change reviewer and specialist lens reads it instead of the full diff; `pattern-triage` keeps the full diff since classification is its job. These ship together because the same CLI writes both artifacts.

No trigger config or consumer files touched.

cc @jeresig

## Testing

- `npx vitest run workflows/review`: 452 tests green, including the new end-to-end smoke case for the provenance gate.
- New unit suites: `lib/diff.test.ts`, `lib/provenance.test.ts`, plus `renderPreExistingNote` rendering tests and router `generatedFiles` coverage.
- `npx tsc --noEmit` clean under the repo's strict compiler options.


workflows/review/lib/render-comment.ts
```diff
@@ -265,6 +265,58 @@ export const renderReviewBody = (input: ReviewBodyInput): string => {
...
+ * carry a blocking label and does not post as an individual comment; instead
+ * all such observations collapse into this one `note (non-blocking)` body.
+ *
+ * Same determinism split as {@link renderComment}: CODE owns the label, the
+ * intro line, the `<details>` wrapping, and each entry's location token; the
+ * MODEL owns each entry's prose (`model_authored_prose`, copied verbatim).
+ * Returns `null` when there is nothing to note; the caller then posts no
+ * note at all.
+ */
+export const renderPreExistingNote = (
```
jeresig: I'm curious if we want to render this at all?

jeresig (approved):
Great - hope this helps to clean up some of the extraneous comments!

workflows/review/lib/render-comment.ts
jwbron: Good call. By the time a finding lands in this bucket it has already slipped past two layers of instruction telling the model not to produce it, so what collects here is mostly drift, and a comment whose body says "no action required in this PR" is exactly the extraneous-comment problem. It would also repeat on re-reviews (the gate runs before the new-scope dedup).

Updated: the note is gone entirely. The gate still sets these findings aside, but they now go to the run artifact (out/pre-existing.json) instead of the PR, so we can still inspect what the gate is catching while posting nothing. A genuinely important pre-existing issue still has a path in: the materially-amplifies rule anchors it on changed lines and it passes the gate as a normal finding.

PR #229, opened 2026-07-09
get-changed-files: fix core.warn crash on multi-PR push events

jwbron:
## Summary

`actions/get-changed-files` crashes with `core.warn is not a function` on push events when the pushed commit is associated with more than one open PR. The github-script `core` API method is `core.warning`; `core.warn` does not exist, and the test mock stubbed `warn` so the suite never caught it. The path fires exactly for stacked PRs (a commit that is the head of one PR and part of another), which is how #227 surfaced it: its "Testing with 'push' events" check dies on this line instead of warning and proceeding with the first associated PR.

One-line fix plus the matching test-mock rename.

cc @jeresig

## Testing

- `npx vitest run actions/get-changed-files`: 40 tests green, including the multi-PR warn path.
- `pnpm build` (tsc) compiles clean.

### DIFF
--- a/.changeset/get-changed-files-warning.md
+++ b/.changeset/get-changed-files-warning.md
@@ -1,0 +1,5 @@
+---
+"get-changed-files": patch
+---
+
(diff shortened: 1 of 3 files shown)


jeresig (approved):
Ahh - good catch!

PR #232, opened 2026-07-09
review: add the live A/B eval plan

jwbron:
Adds `workflows/review/eval/live-ab-plan.md`: a phased, agent-executable plan for real-model evals of review-bot changes, making the round-two doc's eval playbook executable for its behavior-shifting and cost-affecting change classes (today only the deterministic replay exists, which is structurally blind to prompt and model changes).

The phases:

1. **Live-enabled corpus cases**: real diffs, post-change file trees, PR context, and labeled defect specs (live runs choose their own finding ids, so ground truth matches on anchor window + mechanism, not id). Implementation is up in a separate PR.
2. **Live producer**: run the actual sub-agent roster from `review.md` over a case via the Agent SDK (Read/Grep/Glob only, staged checkout, per-agent model pinning, live claim-validator, cost capture), behind the `produceFindings` seam `runner.ts` already ships.
3. **A/B runner + delta report**: same-run arms (merge-base `review.md` vs working tree), five-metric deltas, judge quality deltas, per-case cost, adversarial hard gate outright.
4. **CI wiring**: `review-eval-ab.yml` on `paths: [workflows/review/**]`, sticky PR comment, smoke-live subset by default, `full-eval` label escalation.

Settled decisions recorded in the doc: same-run A/B over stored baselines; candidate corpus + lib for both arms (isolate the model seam); bounded agentic sub-agents, not one-shot completions; report-only except the adversarial gate; budget caps that still emit a partial report. Also records how this instrument relates to seeded live-PR trials like Khan/webapp#40678 (complementary layers, different cadences).

## Test plan:

Docs only (plus an empty changeset). `pnpm run test --run` stays green.

## Next steps (human)

1. Review the settled design decisions section; they bind the whole stack (#233-#238), so disagreement is cheapest to voice here.
2. Two decisions the plan deliberately leaves to humans, no code blocked on them: ratify the gate thresholds that would turn the A/B report into a merge gate (until then everything is report-only except the adversarial rule), and decide whether `pattern-triage` joins the live roster.
3. This PR is docs plus an empty changeset; it can merge independently of the stack whenever the plan reads right. All five phases have landed and the doc now records the phase 4 acceptance results (both arms passed; the plan's acceptance section names the runs and the judge-direction lesson), so the remaining human items are exactly the two decisions in step 2 plus the consumer-repo trial that accepts #238.


jwbron:
Heads up: the eval chain (#233 -> #234 -> #236 -> #237 -> #238, plus #235) was cut from a main that predated review@1.4.0 (missing #228/#230), so its review.md lacked the v1.4.0 disciplines. All chain branches are now refreshed with main via bottom-up merges (no history rewrites), and both acceptance arms (#239/#240) got the refreshed base merged into their heads so the noise-floor and recall-signal measurements characterize the v1.4.0 reviewer that actually runs in production. #240's weakening edit was verified to still isolate to the correctness reviewer against the v1.4.0 text. The refresh pushes will re-trigger the Review Eval A/B runs on both arms; treat the new reports as the acceptance record.

jwbron:
## Reviewer guide for the eval + tuning stack (#232-#252)

Sixteen PRs, one tree. Review and merge bottom-up in three waves; each PR's live A/B report (sticky comment) prices its own marginal review.md delta only, so absolute numbers do not compare across PRs. Full context: the tuning memo (rev 3) and `workflows/review/eval/README.md` (lands in #252).

### Wave 1: eval infrastructure (merge first)
Order: #232 (plan doc) and #233 (corpus format + 10 live cases) from main, then #234 (producer/staging) -> #236 (A/B runner) -> #237 (per-PR CI workflow) -> #238 (review-trial skill); #235 (seeded-trial case port) follows #233 anytime.
**What to review hard:** #233's corpus/loader invariants (case format is the long-term contract) and #236's budget/gate semantics (hard USD caps, adversarial gate, report-not-die degradation). #237 is what spends money on every future PR touching workflows/review; check its triggers and labels. #232 is a doc; skim.

### Wave 2: reviewer tuning, batch two
Order: #243 -> #244 -> #245 -> #247, then two forks off #247: (a) #248 -> #249 -> #250, (b) #246.
**What to review hard:** these are prompt/behavior changes to the production reviewer. Every one was A/B'd (16 arms, $84.59) and the cumulative stack-vs-main run (3x full corpus, $42.83) came back recall-neutral: 21/24 vs 22/24 pooled specs, 28/30 verdicts both arms, zero true misses on the stack arm. Read each PR's per-case regression list and drop buckets, not the percentage rows (single-run deltas below the measured noise floor are decoration). #249 changes spend semantics (budget clamp + landing reserve); #246 adds the re-review mode dial (default unchanged: full).

### Wave 3: eval corpus + measurement (top of stack)
Order: #251 -> #252.
**#251** adds re-review (open-PR) coverage: golden lifecycle cases + the rereview sweep. It sits above #244/#246 because its cases exercise their machinery.
**#252** turns the instrument from tripwire into measurement: repeat aggregation with Wilson intervals (validated by reproducing the cumulative run's hand-computed numbers exactly), `--repeats` powered runs (strict-majority gate; NOTE: more permissive than single-run best-of-three, flagged in the README's statistical-honesty section; push back if you want fail-any), the weekly drift workflow (report -> job summary + artifact + visibility PR under .github/review-eval/drift/), the sql-index spec fix (diagnosed from 28 recorded arm-runs: anchor-site preference, not recall; specs gain altLocations), the Haiku match arbiter (capped, audited via 'fallback', refuse-biased but uncalibrated), ruler-provenance stamps on every report, and the measured noise floor rendered as data (v1, case-asymmetric; see caveats).
**What to review hard in #252:** live-match altLocations semantics, the arbiter's guard rails, the majority-gate policy, and the drift workflow's write permissions.

### After merge: one supervised drift run
Run `gh workflow run review-eval-drift.yml -R Khan/actions` once (~\$60-65, hard-capped at \$85, ~2h). Two reasons: (1) the workflow's final step (committing the report under .github/review-eval/drift/ and opening a visibility PR via GITHUB_TOKEN) has never actually executed, and token pushes can trip branch-protection rules; better to watch it fail than find out on a weekend. Green run + a PR titled 'review: eval drift report <date>' means the flow works. (2) The noise-floor bands hard-coded in the report footer (MEASURED_NOISE_FLOOR in workflows/review/eval/live-ab-report.ts) came from a budget-skipped run and are labeled as inflated; this run clears the full corpus with the arbiter active, so copy its Noise floor table (min/max/sd + run id/date in the provenance string) into that constant in a small follow-up PR.

Known-open (memo, not this stack): anchor-snap (top priority, priced at ~$29 via targeted repeats), arbiter calibration, baseline-arm reuse (blocked on Actions cache scoping), claim-validator flake.

workflows/review/eval/live-ab-plan.md
```diff
@@ -0,0 +1,451 @@
+# Plan: live A/B evals for review-bot changes
+
+Status (2026-07-09): this plan doc is Khan/actions#232. Phase 1 is
```
jeresig: If we're planning on having this be a long-lived artifact then we should probably cleaned up a lot of the PR references and such. But if this only exists for this specific eval then this is fine.

jeresig (approved):

workflows/review/eval/live-ab-plan.md
jwbron: It has ended up long-lived: the eval code and the review-trial skill cite it by phase name. Cleaned up in 8beeeb6; the status header is now a stable intro, and the remaining PR references read as historical citations rather than live status.

PR #237, opened 2026-07-09
review: per-PR live A/B workflow (phase 4)

jwbron:
Phase 4 of the live A/B eval plan (#232), stacked on the Phase 3 runner (#236): the `Review Eval A/B` workflow.

## Behavior

- **Triggers**: every non-draft PR touching `workflows/review/**` (opened/synchronize/reopened/ready_for_review), plus `workflow_dispatch` with `base_ref`, `max_usd`, and `full` inputs.
- **Runs** `live-ab.ts` against the PR's merge-base (`origin/<base>`; `fetch-depth: 0` so `git show` can read the baseline `review.md`), posting the delta report three ways: a sticky PR comment (hidden-marker upsert, edited in place on later pushes), the job summary, and an uploaded artifact (`if: always()`, so a partial or gate-failing run still surfaces what it wrote).
- **Scope and cost**: per-PR runs cover only live cases also tagged `smoke` (the runner gains `--smoke-only`); a `full-eval` label or the dispatch input lifts that to all live cases. The runner's own budget cap (default $40 total, half per arm) reports skipped cases rather than dying. Per-PR concurrency cancels superseded runs on a new push.
- **Opt-outs and guards**: `skip-live-eval` label skips entirely; drafts wait until ready-for-review; the `changeset-release/main` branch is excluded (it matches the path filter via `workflows/review/package.json` but changes no behavior); a missing `ANTHROPIC_API_KEY` (fork PRs) skips green, same pattern as `review-eval-full.yml`.
- **Gate**: report-only except the runner's exit code, which is non-zero only when the candidate arm mishandles an adversarial-injection case (the playbook's standing rule). Metric deltas never fail the job until thresholds are ratified (plan's open human decision).

The workflow name is distinct from every gh-aw workflow per the operational-floor rule (same-named workflows share a gh-aw concurrency group per PR and cancel each other).

## Test plan:

- YAML parses; `pnpm run test --run` (696) and typecheck/lint stay green with the `--smoke-only` and markdown-sibling additions to `live-ab.ts`.
- Real validation requires the repo secret: after this stack merges (or via `workflow_dispatch` on the branch), the acceptance runs from the plan are (1) a PR editing only a `review.md` comment produces a near-zero-delta report comment, and (2) a PR deliberately weakening a reviewer instruction shows a recall regression. Those two runs should be linked here before undrafting.

## Next steps (human)

1. ~~Labels~~ DONE: `full-eval` and `skip-live-eval` exist in this repo (created 2026-07-09).
2. ~~Secret~~ DONE: confirmed working by the acceptance runs below.
3. ~~Acceptance~~ DONE and PASSED, run from the two (now closed) do-not-merge PRs #239 and #240: the control (comment-only edit, run [29052332224](https://github.com/Khan/actions/actions/runs/29052332224)) held recall and verdict agreement flat at +0% with the adversarial gate passing, and the weakened-reviewer arm (run [29052334404](https://github.com/Khan/actions/actions/runs/29052334404)) showed recall -17% with the lost spec named (`incident-money-rounding:money-fp-rounding-1`) and verdict agreement -14%. ~26 min and ~$7.70 per full run. Details are in the plan's acceptance section (#232) and the evidence comment below.
4. Decide the default budget: `$40` total per run is the shipped default; at a few review-bot PRs a week that is bounded, but it is your spend to ratify.
5. Cross-check workflow-name distinctness once more against any consumer-repo gh-aw workflows that could run on this repo's PRs (the shipped name `Review Eval A/B` collides with nothing in Khan/actions today).

---

## Update 2026-07-09: baseline semantics documented

The workflow passes the base-branch **tip** while its prose said merge-base. The code was right: pull_request runs check out the PR merge commit, so the candidate tree already contains the base tip, and baselining on that same tip isolates the PR's own review.md delta (a merge-base baseline would fold upstream review.md movement into the measured numbers). Prose updated to match; no behavior change.


jwbron:
**Both acceptance runs from the test plan are done and PASSED**; this satisfies next-steps item 3.

- Control (#239, run 29052332224): identical prompts, zero recall/verdict delta, adversarial gate passed. Measured noise floor: about 5 points on the noise metric, about 0.1 on judge mean quality, roughly $7.75 and 13 minutes per PR run (both arms, 7 smoke-live cases).
- Signal (#240, run 29052334404): weakened correctness reviewer detected exactly as designed: recall 83% to 67% with the lost spec named, verdict agreement 100% to 86%, adversarial gate still passing. Notably, judge mean quality ROSE on the regressed arm (fewer, surer comments score better per comment): the recall-against-specs metric is load-bearing, judge quality alone is not.

The runs also validated the two earlier fixes live (case-namespaced finding ids; judge failures degrading instead of killing the report) and surfaced two follow-ups:

1. `incident-sql-missing-index:dm-missing-index-1` was missed by all four arms. The case's recorded finding is a data-migrations LENS finding, but the converted live cases carry no `routerConfig` lens rules, so the live roster never spawns the specialist. Fix: add per-case `routerConfig` lens rules to the live incident cases mirroring the lens each recorded finding came from (a #233-corpus data change).
2. The claim-validator failed on `incident-cache-missing-key` in the baseline arm of BOTH runs, and the report's per-case table only names the failed agent, not its `failed` reason. Fix: carry the failure message into `perCase` so the next occurrence is diagnosable from the report alone (a #236 change), then investigate the validator failure itself.

.github/workflows/review-eval-ab.yml
```diff
@@ -0,0 +1,133 @@
...
+# `skip-live-eval` label skips the job entirely.
+#
+# NOTE: this workflow's name must stay distinct from any gh-aw workflow a
+# consumer runs (same-named workflows share a gh-aw concurrency group per PR
+# and cancel each other; see the round-two doc's operational floor).
+
+name: Review Eval A/B
+
+on:
+  pull_request:
```
jeresig, line 32: So it seems like we're running this once per PR here in this repo? (maybe twice if we start as a draft?)

jeresig (approved):

.github/workflows/review-eval-ab.yml
jwbron: Slightly more than that: it runs on every push to a non-draft PR touching workflows/review/**, with cancel-in-progress superseding a stale run on a new push. Drafts skip entirely until ready_for_review, so starting as a draft does not add a run. Each run is the smoke subset (7 live cases per arm), measured at ~$7.70 and ~26 minutes in the acceptance runs; skip-live-eval opts a PR out, full-eval widens to the whole live corpus, and the runner enforces a hard USD ceiling either way. If per-push proves too spendy we can gate on a label instead, but per-push is what makes it a tripwire between a prompt edit and production.

PR #239, opened 2026-07-09
[do-not-merge] live A/B acceptance: control (comment-only)

jwbron:
**DO NOT REVIEW, DO NOT MERGE.** This PR exists only to trigger the `Review Eval A/B` workflow (#237) as its acceptance CONTROL; it will be closed unmerged once the report lands.

The only change is one period added to a frontmatter comment in `review.md`. No extracted agent prompt contains that line, so both A/B arms run byte-identical prompts: the report's deltas measure pure run-to-run model variance, which is the noise floor to record before interpreting any real A/B report.

Acceptance criteria (from the plan, `workflows/review/eval/live-ab-plan.md`):
- The runner logs the identical-arms note.
- Near-zero metric deltas and an empty regressions list in the sticky comment below.
- Whatever deltas remain = the measured noise floor; note it in #237 before undrafting the stack.

Companion signal PR: the deliberately-weakened-reviewer acceptance test.


jwbron:
Companion signal PR: #240.

jwbron:
**Acceptance: PASSED.** With byte-identical prompts in both arms, recall and verdict agreement deltas are exactly zero, no clean false flags, and the adversarial gate passed. The measured run-to-run noise floor, which future reports should be read against: noise metric about 5 points, judge mean quality about 0.1, cost about $3.9 per arm, 12-13 min per arm.

Two observations recorded for follow-up (details in the #237 acceptance comment): `incident-sql-missing-index` is missed by every arm (the case's recorded finding comes from the data-migrations lens, but live cases carry no `routerConfig` lens rules, so the specialist never spawns), and the claim-validator failed on `incident-cache-missing-key` in the baseline arm of both runs while the report only names the agent, not the failure reason.

Run: https://github.com/Khan/actions/actions/runs/29052332224. Closing as planned; do not reopen or merge.

PR #241, opened 2026-07-09
Enable the automatic AI PR reviewer on this repo

jwbron:
## Summary

Enables the shared PR reviewer (`workflows/review/review.md`) on this repo itself, pinned to `review-v1.4.1`. Unlike webapp's `/review` slash-command setup, this install keeps the shared prompt's automatic `pull_request` trigger: every non-draft PR push gets reviewed, with the `skip-ai-review` label (created, matching webapp's) as the per-PR opt-out.

### Installed workflow (`review.md` / `review.lock.yml`)

`gh aw add` refuses same-repo installs, so the prompt is copied from the `review-v1.4.1` tag with a `source:` pin recorded, then compiled with `gh aw compile` (v0.81.6). Local overrides on top of the shared source, all marked `KHAN/ACTIONS LOCAL OVERRIDE` inline:

- **Fork-PR guard**: `github.event.pull_request.head.repo.full_name == github.repository` prepended to the job-level `if:`. This repo is public (webapp is private); with the shared prompt's `roles: all` disabling gh-aw's own actor gate, this condition is what keeps untrusted fork heads from triggering the workflow. Same-repo branches require write access to create.
- **Lib checkout ref**: comment-only as of v1.4.1. The tag's own `ref:` line now matches the release (#242/#253 wired the sync into the release flow), so no hand-bump is needed; the inline comment keeps the self-hosting rationale that the reviewer runs the released lib, not the PR head; a PR must not be able to change the code that reviews it.

### Consumer config (`.github/aw/review/`)

- **`config.md`**: `add-reviewer` allowlisted to `github-actions` (the sole owning team per `.github/REVIEWERS`), `max: 2`, using `KHAN_ACTIONS_BOT_TOKEN`.
- **`risk-classification.md`**: written for this repo's supply-chain blast radius; the composite actions and shared workflows execute inside consuming repos' CI with their credentials. High tier: `actions/**`, the publish pipeline (`utils/`, `release.yml`), `.github/workflows/**`, and the reviewer prompt + lib. Deep-checks cover template injection in `action.yml` `run:` blocks, action SHA-pinning, publish/tag integrity, gh-aw frontmatter security surfaces, `.md`/`.lock.yml` pairing, and semver-as-behavior-contract on `review` releases.
- **`ci-tooling.md`**: what node-ci already catches (eslint/prettier, tsc, vitest, the action smoke tests, changeset presence) so the reviewer stays silent on those.
- **`skills.md`**: this repo has no `.claude/skills/`; the catalog points the skill-auditor at the two real convention docs (root `README.md` for the release/changeset flow, `workflows/review/README.md` for the reviewer contract) and forbids flagging conventions from memory.
- **`ROUTING`**: enables the same eval-justified whole-change roster as webapp (`holistic,completeness,test-adequacy,first-principles,conventions`); no specialist lenses, since all eleven target webapp's backend domains. Tier rules verified against the lib parser (`lib/routing-config.ts`): parses with zero warnings.

### Supporting changes

- **`.gitattributes`**: `gh aw compile` itself added `.github/workflows/*.lock.yml linguist-generated=true merge=ours`.
- **`node-ci.yml`**: `.gitattributes` added to the `check-for-changeset` exclude list, alongside the lint configs already there; metadata changes should not force cutting a release. No changeset on this PR for the same reason (everything else is under `.github/`, already excluded).

## Security review note (gh-aw compile)

- **Secrets**: the new lock's manifest lists `ANTHROPIC_API_KEY`, `GH_AW_GITHUB_MCP_SERVER_TOKEN`, `GH_AW_GITHUB_TOKEN`, `GH_AW_OTEL_SENTRY_AUTHORIZATION`, `GH_AW_OTEL_SENTRY_ENDPOINT`, `GITHUB_TOKEN`, `KHAN_ACTIONS_BOT_TOKEN`; the identical set webapp's production `review.lock.yml` uses. Availability here: `ANTHROPIC_API_KEY` exists at repo level (review-eval-full already uses it); `KHAN_ACTIONS_BOT_TOKEN` is org-level (gerald-pr and release already use it); the MCP tokens fall back through `|| secrets.GITHUB_TOKEN` chains, so their absence is fine. The two `GH_AW_OTEL_SENTRY_*` secrets are not configured in this repo, and that turned out to be a hard failure, not a graceful degrade: the MCP gateway's OTLP config schema rejects the empty endpoint URL and the agent job dies at startup (observed on this PR's first run). The `observability:` block is therefore disabled as a marked local override, removing both secrets from the lock's manifest; restore it verbatim once the two secrets exist here.
- **Actions**: all pinned to the same SHAs as webapp's production lock (`actions/checkout`, `cache`, `github-script`, `setup-node`, `upload/download-artifact`, `github/gh-aw-actions/setup@v0.81.6`); no actions new to Khan's review workflows.
- **Triggers/permissions**: `contents: read` + `pull-requests: read` at the job; all writes go through gh-aw safe outputs with the caps from the shared v1.4.1 prompt (reviewed in #226/#227/#228/#231). The only trigger-surface change relative to webapp is the fork guard described above, which tightens, not widens.

## Testing

- `gh aw compile` clean (0 errors, 0 warnings); the lock carries the fork guard and the `github-actions` allowlist (verified by inspection of the compiled `if:` and safe-output handler config).
- `ROUTING` parsed with the repo's own `workflows/review/lib/routing-config.ts`: expected tiers/roster, `warnings: []`.
- The workflow runs on this PR itself (a `pull_request` workflow executes from the merge ref, so a PR that adds one exercises it). The first push's activation failed because `risk-classification.md` quoted literal template-expression syntax in its injection deep-check; gh-aw validates runtime imports and rejects that. Rewritten in prose. The second run then failed at agent startup on the missing OTLP secrets (see the security note); the observability block is now disabled locally. The run on the current push is the live smoke test.


.github/aw/review/ROUTING
```diff
@@ -0,0 +1,33 @@
...
+
+# Dev-only surfaces.
+workflows/review/eval/**     tier=low
+**/*.test.ts                 tier=low
+
+# Docs are trivial; agentic-workflow .md files are executable, not docs, so they
+# are re-raised after the broad docs rule (last match wins).
+**/*.md                      tier=trivial
+.changeset/*.md              tier=trivial
+workflows/review/review.md   tier=high
```
jeresig: I think we want this to match any workflow .md file (I'm sure we'll have others)

.github/aw/review/risk-classification.md
```diff
@@ -0,0 +1,131 @@
+<!--
```
jeresig, line 1: Fine for now - I'm curious how much of this file we still need wit the new ROUTING classifier (probably this should only be relevant for files that weren't previously classified by ROUTING)

jeresig (approved):
Great - looks good to me!

.github/aw/review/ROUTING
jwbron: Done in f3b902f: `workflows/*/*.md` is now tier=high, with the co-located README/CHANGELOG dropped back to trivial. Single-star segments don't cross into subdirectories, so nothing under eval/ is re-raised. Re-validated against `lib/routing-config.ts`: zero warnings, tiers unchanged for every currently checked-in path.

.github/workflows/review.md
```diff
@@ -0,0 +1,3085 @@
...
+      ref: review-v1.4.0
+      path: gh-aw-review-lib
+      persist-credentials: false
+
+# Cost guardrails (AI credits; 1 credit = $0.01). gh-aw >= v0.79 bakes in
+# defaults of 1000/run ($10) and 5000/day ($50). Disable the daily ceiling
+# (-1) so reviews are never skipped on a busy PR day; the per-run cap below
+# still bounds the cost of any single review.
+max-daily-ai-credits: -1
+source: Khan/actions/workflows/review/review.md@review-v1.4.0
```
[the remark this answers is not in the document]
jwbron: Addressed the cheap half in 13f143e: `.github/workflows/review-pins.test.ts` fails any PR where the `source:` ref and the lib checkout `ref:` in the installed file differ (the shared source's own ref is covered separately by #242's sync + backstop). Skipped the fuller check (diffing the installed file against the tagged source with an override-hunk allowlist); it needs release tags fetched in CI and feels disproportionate for now.

.github/workflows/review.md
```diff
@@ -0,0 +1,3085 @@
...
+# GH_AW_OTEL_SENTRY_AUTHORIZATION — the `sentry sentry_key=<public-key>` header value.
+#
+# KHAN/ACTIONS LOCAL OVERRIDE: the shared source's `observability:` block is disabled
+# here because this repo has no GH_AW_OTEL_SENTRY_* secrets configured, and a missing
+# endpoint is NOT a graceful degrade: the compiled lock feeds the empty value into the
+# MCP gateway's OTLP config, whose schema requires a non-empty https:// URL, so the
+# agent job fails at startup (observed on the first run of PR #241). Restore the block
+# below verbatim once the two secrets exist in this repo.
+#
+# observability:
```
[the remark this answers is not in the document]
jwbron: Documented upstream in #254: the two secrets are now in the README's required-secrets section, and the comment on the `observability:` block itself carries the same warning plus the workaround (comment the block out as a preserved local edit and recompile). A skip-when-empty would need gh-aw support, so prose for now.

.github/aw/review/risk-classification.md
jwbron: Agreed on the destination. ROUTING should be the single source of truth for path-to-tier, with this file refocused on what globs can't express (deep-checks, direction-dependent judgments, unclassified paths). I'd hold off on a shared-prompt change for now though. Router tiers already win where it counts (budget and verdict), ROUTING is optional while this file is required, and taking tiers as given is a behavior change per the semver contract, so it wants an eval run once the A/B tooling lands. In the meantime this repo's file already leans that way (mostly deep-checks, thin tier prose), and we can treat that as the convention for new installs.

PR #242, opened 2026-07-09
review: keep the pinned review.md ref in sync with the released version

jwbron:
## Problem: the ref: lag

`workflows/review/review.md` checks out Khan/actions at a pinned tag in its `pre-agent-steps` to fetch the deterministic lib the prompt invokes at runtime. The file's own comment calls that `ref:` "the single version surface for prompt + code", but nothing updated it during a release: changesets bumps `package.json` and changelogs, not prose in a markdown file, so every tag since review-v1.2.2 (v1.3.0, v1.3.1, v1.4.0) shipped with the ref still reading `review-v1.2.2`.

The skew is not cosmetic: the v1.4.0 prompt invokes `lib/provenance.ts`, which does not exist in a v1.2.2 checkout, so an un-overridden consumer gets runtime failures or silently missing gates. It also mis-stamps attribution, since the posted `pr-reviewer:version` marker reads the version from the pinned checkout's `package.json`. (This is the "operational floor" row from the round-two plan, re-confirmed on the review-v1.4.0 tag on 2026-07-09.)

## Fix: automation, not vigilance

- **utils/sync-workflow-versions.ts** (+ lib, tests): generic across workflows (per review feedback). Discovers every directory under `workflows/` (release tags are named after the directory, matching `publishWorkflow`) and rewrites every `<name>-v<semver>` literal in that workflow's markdown files (for review, today, only the checkout `ref:` line) to that workflow's package version. CHANGELOG.md is skipped (historic versions live there on purpose); a workflow with no literals is a no-op.
- **release.yml / package.json**: the changesets action now runs `pnpm run version-packages` (`changeset version` followed by the sync) instead of the default `changeset version`, so the ref bump lands in the same Version Packages commit that gets tagged. If a Version Packages PR does not bump any workflow package, the sync is a no-op.
- **workflows/review/version-sync.test.ts** (CI backstop): fails any PR, including the Version Packages PR, where review.md's literals do not match the `review` package version. Runs with the existing `pnpm run test --run` step in node-ci.
- Bumps the currently stale ref to `review-v1.4.0` so the invariant holds on main now; the next release rewrites it automatically.
- Doc updates: review.md's version-surface comment now describes the sync mechanism, and the README's install/versioning sections state that each tag's review.md pins its own release (repo-wide audit of `review-v` literals found no other embeds outside changelogs and test fixtures).

Changeset included (patch bump for `review`; release plumbing and comments only).

## Verification

- Simulated the release end to end by running `pnpm run version-packages` against this PR's real changeset (re-run after the generic rewrite): `changeset version` bumped `review` to 1.4.1 and the sync rewrote the ref to `review-v1.4.1` in the same working state the changesets action commits (then reset).
- Negative check: with the ref set back to `review-v1.2.2`, the backstop test fails as intended.
- `vitest` suite, `eslint`, and `tsc` all green on the real tree.

Once a release ships with the sync in place, webapp's hand-pinned override (its imported review.md carries a "the tag's own ref line lags the release" comment and hand-bumps the ref on every re-import) can be dropped on its next re-import.

cc @jeresig

## Human steps (non-review)

- [x] **Verify on the next release**: done 2026-07-10. #253 released `review-v1.4.1` and the tagged `review.md` carries `ref: review-v1.4.1`; the ref bump landed in the same Version Packages commit that was tagged.
- [x] **Retire the consumer workaround**: Khan/webapp#40756 re-imports at `review-v1.4.1` and drops the hand-pinned `ref:` override comment. Khan/actions#241 (self-hosted install) was bumped to `review-v1.4.1` the same way.


.github/workflows/release.yml
```diff
@@ -30,6 +30,11 @@ jobs:
         if: steps.has-changesets.outputs.result == 'true'
         uses: changesets/action@6a0a831ff30acef54f2c6aa1cbbc1096b066edaf # v1
         with:
+          # Not the default `changeset version`: version-packages also syncs
```
jeresig:
```suggestion
          # Note the default `changeset version`: version-packages also syncs
```

utils/sync-review-version-lib.ts
```diff
@@ -0,0 +1,79 @@
+import * as fs from "fs";
```
jeresig: Could we make this generic so that it updates any workflow? We're bound to have more shared workflows so we might as well handle this in a single way.

jeresig (approved):
Thanks for solving this issue - one request inline to make this generic for any possible workflow that we add!

utils/sync-review-version-lib.ts
jwbron: Done in 3eafaab: the script is now utils/sync-workflow-versions.ts. It discovers every directory under workflows/ (the tag is named after the directory, matching publishWorkflow in utils/publish.ts), reads its package.json version, and rewrites every `<name>-v<semver>` literal in that workflow's markdown files, skipping CHANGELOG.md since changesets writes historic versions there on purpose. A workflow whose markdown embeds no literal is a no-op; a workflow that must embed one (like review.md's checkout ref) keeps its own contract test as the CI backstop.

.github/workflows/release.yml
jwbron: Reworded this comment as part of making the sync generic; it now leads with what version-packages does rather than the "Not the default" phrasing.

PR #244, opened 2026-07-09
review: re-review accountability, code-rendered from the reconciler keep-list

jwbron:
## Summary

Bottom of the fold-in batch two stack, now based on `main` (#232-#238 and #243 merged 2026-07-10; the branch has since merged main, so the files view shows only this PR's changes). **The top of the batch-two stack is `jwbron/review-skill-rule-quote` (#248).**

Today a re-review resolves the threads the author fixed and says nothing about the rest. Observed on the v1.4.0 re-run lifecycle (Khan/webapp#40730): run 2 left three blocking threads open and unacknowledged under a bare "Changes requested" body, and run 3 approved with an empty body while resolving 11 threads and posting 8 notes. The author reading the verdict has no way to see which earlier findings remain open.

Failure scenario fixed: an author pushes a partial fix; the re-review resolves two of five threads and requests changes with an empty body; the author reasonably assumes everything visible is new, misses the three open blocking threads (one is collapsed as outdated), and pushes again without addressing them; the cycle repeats.

The verdict body now accounts for every prior thread, and the section is code-rendered rather than prompt-trusted:

- New `lib/rereview.ts`: renders the accountability section deterministically from the `thread-reconciler`'s keep/resolve lists. Still-unaddressed threads are enumerated as links to the prior comments ("N of M prior review threads resolved; K still unaddressed as of \<sha\>:"), blocking first, each with the prior comment's Conventional-Comment label and a verbatim, code-truncated excerpt of its opening line (text this workflow itself posted on the earlier run; no new model prose). When the run resolved the last open threads, the section states "All N prior review threads are resolved.", so an approval accounts for its resolutions. Fail-open: missing or unparseable staging inputs render an empty section, leaving the body exactly as today.
- `threads.json` staging gains `url` (the thread's first comment `html_url`, already present in the `get_review_comments` output) to power the links; a thread without one renders as a plain `path:line` token.
- `renderReviewBody` gains a `rereviewSection` slot spliced between the verdict head and the note lines, and review.md Step 6 tells the orchestrator to run the CLI and append its `section` verbatim, never to compose the accounting itself. The redundant-approval skip treats a non-empty section as content, and since resolved threads leave `threads.json` on later runs, the "all resolved" line appears only on the run that resolved them (no repeat-approval noise).

Scope cut, deliberate: the optional batched "still open as of \<sha\>" reply per still-open blocking thread is not implementable on gh-aw v0.81.6; there is no reply-capable safe output (`create_pull_request_review_comment` has no in-reply-to parameter, checked against the v0.81.6 tool schema), and a fresh inline comment at the same path:line would open a duplicate thread, which the reconciler dedup rules exist to prevent. The verdict-body enumeration carries the accountability.

cc @jeresig

## Testing

- `npx vitest run workflows/review`: 477 tests green (20 new in `lib/rereview.test.ts`: label parsing, blocking-first ordering, singular/plural and all-resolved wording, link and no-url fallbacks, body splicing, CLI fail-open paths).
- `npx tsc --noEmit` clean.


jwbron:
Branch rebuilt on the refreshed chain (491a983, content unchanged; see #243 for the full topology note). Reset local state to origin before continuing.

jeresig (approved):
It looks like this PR is now including a bunch of additional changes? I looked at the re-review stuff and that makes sense to me.

jwbron:
Good catch: the extra changes were an artifact of a stale merge base. The PR diff was being computed against main as of #233, so the whole live-eval stack that has since landed on main (#235-#237, #243) showed up in the files view here. The actual change is only the re-review accountability work: lib/rereview.ts and its tests, the render-comment slot, and the review.md step. I've merged main into the branch and the files view now shows just that (6 files, ~680 lines).

PR #245, opened 2026-07-09
review: hand off out-of-lane observations instead of dropping them

jwbron:
## Summary

Third of five PRs in the fold-in batch two stack (stacked on #244; the stack is based on #238, `jwbron/review-trial-skill`). **The top of the batch-two stack will be `jwbron/review-skill-rule-quote`.**

In run 2 of the v1.4.0 re-run lifecycle the skill-auditor found a real issue (dedup reads via an eventually-consistent Query immediately after PutMulti) and dropped it, verbatim: "that's a correctness concern, not a quotable skill-rule violation, so I'll leave it". That was the correct call under quote-the-rule, but the observation died with it.

Failure scenario fixed: a reviewer investigating a rule notices a genuine correctness defect, correctly declines to report it as a skill violation, and has no other channel; the defect reaches the PR author in no form and ships.

The handoff:

- The skill-auditor and all 11 specialist lenses may now return `out_of_lane_observations[]` alongside `findings[]`: `path`, optional `line`, the concern stated concretely, a required concrete `failure_scenario` (the claim the validator attacks, same discipline as findings), and an optional `suggested_lane`.
- A new orchestrator rule (Step 3, after lens normalization) converts each observation into a candidate comment with the code-assigned label `question (non-blocking)`; the label is never model-chosen, because a handoff is not a vetted finding and must not block on its own (and the claim-validator never upgrades severity). Candidates then flow through the identical provenance gate, scope filter, claims.json, and validation path as every other candidate; validation into the correctness lane, rather than re-dispatching the correctness reviewer, keeps the handoff inside the run's existing budget.
- The shape is code-validated: `validateOutOfLaneObservation` / `isValidOutOfLaneObservation` in `lib/finding-schema.ts`, with the same collect-every-error contract as `validateFinding` so a malformed handoff is diagnosable from the run artifact. It is a sibling type, not a `Finding` field, so `FINDING_SCHEMA_VERSION` stays 2 and no serialized finding is invalidated.

The per-lens handoff paragraph and JSON field are stamped into each of the 11 lens definitions (mechanically, identical text); the next PR in the stack (dispatch-tax trim) folds exactly this kind of repeated block into the shared staged disciplines context, so the duplication added here is transitional within the same stack.

cc @jeresig

## Testing

- `npx vitest run workflows/review`: 484 tests green (7 new observation-schema tests: minimal and full shapes, required fields, error collection, line bounds).
- `npx tsc --noEmit` clean.


jwbron:
Branch rebuilt on the refreshed chain (7b5318c, content unchanged; see #243 for the full topology note). Reset local state to origin before continuing. The dispatch-tax trim PR should base on this branch.

jeresig (approved):
Looks good!

PR #246, opened 2026-07-09
review: the re-review mode dial (full | scoped | flip-gated | fast)

jwbron:
The re-review mode dial: the runs-per-PR cost lever from the plan of record's "Re-review economics" section. Stacked on fold-in batch two (#244 -> #245 -> #247, base `jwbron/review-dispatch-tax`); the eval stack (#232-#238, #243) merged to main 2026-07-10.

## Why

Two measured lifecycles of the same seeded PR paid the full 11-agent fan-out on every push, with per-run cost rising as threads accumulate (v1.3.1: $7.19 / $8.77 / $9.14; v1.4.0: $8.31 / $9.87 / $11.29, runs 29044383251 / 29045581242 / 29046928431 in Khan/webapp). The approval run is the most expensive while emitting the least. Both lifecycles also caught a fresh seeded defect on a re-review push that a reconcile-only fast path would have missed, which argues for `scoped` as the default reduced depth.

## What

A per-repo mode line in `.github/aw/review/ROUTING`:

```
re-review full|scoped|flip-gated|fast
```

- `full`: today's behavior, and the default. **No consumer changes behavior until its repo sets a mode line.**
- `scoped`: full roster, staged only the hunks new since the last fully-reviewed fingerprint; comments scoped to those hunks.
- `flip-gated`: reconcile plus the correctness pass over new hunks; a REQUEST_CHANGES to APPROVE flip is vetoed by that pass's validated blocking findings (the findings gate the flip instead of being discarded).
- `fast`: reconcile-only.

Three guards, all deterministic (`workflows/review/lib/rereview-mode.ts`):

1. **Ready-for-review anchor**: the one full review the cheap modes lean on is taken at ready-for-review, never on a draft skeleton (a draft-taken fingerprint forces one more full review when the PR leaves draft).
2. **Flip gate**: as above; mechanically this rides the existing verdict function, since a validated blocking finding posts and blocks.
3. **Divergence tripwire**: every full-depth review stamps a content-hashed hunk signature into its review body as a hidden comment, so it survives cache eviction. Each push is compared against the last fully-reviewed fingerprint; at unreviewed share >= 0.4 the run re-arms to full. Signatures are computed over the generated-stripped diff, so lockfile churn neither enters the fingerprint nor counts as divergence.

Named interactions handled by construction:

- **dismiss-stale-approvals**: the prior verdict and fingerprint live in the stamp, not the review `state`, so a dismissed approval still anchors and still reads as an approval.
- **COMMENTED-only prior review**: the stamp itself (not the verdict state) marks "a full review happened", so a comment-only history reduces normally after its first stamped run instead of re-running full forever.

## Eval and measurement

The roster-level evals live in #251 (stacked on this PR): open-PR corpus cases with reconciler dispatch and flip-gate/duplicate scoring, the per-mode pricing sweep (`eval/rereview-sweep.ts`), and the ported seeded lifecycle. This PR carries the deterministic layer:

- Adversarial lifecycle cases as data (`eval/lifecycle/`): `rewrite-after-approval` and `sparse-pr-then-payload`, each scored as "the tripwire re-armed and the payload got a full review", plus an economy control (`fix-push-stays-cheap`) scored as "the reduced path actually ran". Replayed deterministically in `eval/lifecycle.test.ts` under all three reduced modes.
- Pricing wiring for the live A/B: each run's executed depth lands in `out/rereview-plan.json`; `lib/counters.ts` aggregates `costByRereviewDepth`; the review-trial skill gains a ROUTING-override arms pattern (two arms at the same ref differing only in the mode line) and scores adversarial lifecycle pushes.
- This is the playbook's cost-affecting class: mode changes are earned per repo through the live A/B on recall and dollars, the way `enable` lines are.

## Composition with batch two

Based on the top of the batch-two stack (#243 -> #244 -> #245 -> #247) so the mode dial's re-review paths compose with the re-review accountability rendering (#244) rather than race it: the accountability section stays the visible body surface; the fingerprint stamp is appended last as a hidden comment; and the flip gate reads `keptBlockingCount`, a small additive field this PR adds to the accountability renderer's `rereview.json`, so "did the reconciler keep any blocking thread" is a number read at verdict time, not a label judgment re-made there. The orchestrator wiring (Step 1 prior-review staging, the Step 3 depth plan and per-depth dispatch/staging, the Step 4 flip rule, the Step 6 stamp, reduced-depth skips in Steps 7/8) is the last commit on the branch.

## Supersedes

Khan/frontend#13420 (frontend's consumer-side re-review stopgap, held on the old pin) is superseded by this: the dial lives in the shared workflow and frontend opts in with a single `re-review` line in its ROUTING file once priced. Frontend is deliberately not touched here.

## Follow-up: staging moves pre-workflow (2026-07-13)

Per review feedback (jeresig): the deterministic staging this PR wires through orchestrator prompt steps (the Step 1 prior-review fetch, the rereview-mode CLI invocation, the scoped-diff swap) needs no model output and will migrate into pre-workflow steps the agent just reads, folded into the round-three deterministic-orchestrator work rather than special-cased here. The eval producer moves with it so the A/B keeps measuring production. This PR's prompt wiring is interim and gets deleted wholesale when that migration lands.


jwbron:
Heads-up: jwbron/review-dispatch-tax (this PR's base) was rebuilt on the refreshed chain per the topology note on #243; its history was rewritten (old tip 8e015ff, new tip 813767c, content unchanged apart from prettier formatting). Reset or rebase local state onto origin/jwbron/review-dispatch-tax before continuing. The prettier/lint CI failures on the old chain are fixed on the new one.

jwbron:
Eval pointer: the real (roster-level) evals for the mode dial live in #251, stacked on this PR.

This PR carries the mechanism plus its deterministic layer only: the depth/tripwire replays in `eval/lifecycle/` (rewrite-after-approval, sparse-PR-then-payload, the fix-push economy control), the unit-level guards, and the cost-attribution surfaces (`out/rereview-plan.json`, `costByRereviewDepth`).

What actually measures reviewer behavior under each dial setting is in #251:

- open-PR corpus cases (`live.rereview`): staged prior threads, the stamped prior review, the depth plan; the live producer dispatches the `thread-reconciler` and sizes the finder roster by depth
- scoring: thread-resolution accuracy, the flip-gate input (kept blocking count), duplicate comments on kept threads, plus fresh-defect recall via the ordinary `mustCatchSpecs`
- pricing: `eval/rereview-sweep.ts` runs all four modes in one command (recall, resolution, flips, dups, dollars, with the EXECUTED depth per case since the tripwire can override the dial), and `live-ab.ts --re-review-mode` prices one mode against a full-depth baseline
- `golden-retention-lifecycle-1/2/3` (sanitized port of the measured webapp#40730 seeded lifecycle) and `golden-retention-fix-push` (the under-tripwire case where the reduced paths actually execute)

Reviewers judging whether the dial is measured before any repo flips a mode line: that evidence is #251's; both merge before any consumer can set `re-review`, since this PR ships default-full.

workflows/review/review.md
```diff
@@ -359,6 +359,15 @@ Write the result to `/tmp/gh-aw/review/new-scope.json` as
 RIGHT-side line numbers of the added lines inside in-scope hunks. Step 3 uses this to
 filter candidate comments.
 
+**Stage the bot's prior reviews.** Fetch the PR's reviews (`pull_requests`
```
jeresig, line 367: Musing: I think we would save a lot of money/tokens/time by moving a lot of this (and similar staging) logic into Github action steps that run before the workflow that downloads and stages everything and then it can just read what it needs, if it needs it.

workflows/review/review.md
```diff
@@ -526,6 +535,53 @@ two files:
   so a lock-file-heavy PR cannot balloon their context; `pattern-triage` still reads
   `full.diff` because classifying every changed file is its job.
 
+**Decide the re-review depth (deterministic code).** After the provenance CLI, run
+the re-review mode CLI from the shared lib checkout, once:
+```
+cd gh-aw-review-lib && npx -y tsx workflows/review/lib/rereview-mode.ts
```
jeresig, line 546: If we did the pre-run we could run these scripts ourselves and then just feed in the results in here (which would also improve speed and reduce token usage).

workflows/review/review.md
```diff
@@ -526,6 +535,53 @@ two files:
...
+```
+It reads `routing.json` (the repo's `re-review` mode line, default `full`),
```
jeresig, line 548: I think some of this context isn't needed, the reviewer here only cares about what the file is, not how it arrived at that solution.

workflows/review/review.md
```diff
@@ -526,6 +535,53 @@ two files:
...
+overflow or a missing input forces `full`, and the divergence tripwire re-arms
+`full` when too much of the diff is unreviewed. What each depth means for the phases
+below:
+
+- **`depth: full`**: proceed exactly as written below; nothing changes.
+- **`depth: scoped`**: the full roster runs, but over only the unseen hunks. Before
+  Phase 1, overwrite `/tmp/gh-aw/review/full-stripped.diff` with the contents of
+  `scoped.diff`, and in Phase 1 build `pr.diff` from the `scoped.diff` sections of
+  the triage `reviewFiles` (a `reviewFiles` entry absent from `scoped.diff` is
+  already reviewed; leave it out of `pr.diff`). Everything else, the provenance
```
jeresig, line 568: Again, this feels like stuff that we could just set up ahead of time so that it can't ever get it wrong.

jeresig (approved):
Great! Pushing back a bit so that we move more of this staging logic into a pre-workflow step, rather than relying on the agent to run this.

jwbron:
Agreed, and this is where round three is headed: the deterministic-orchestrator work covers exactly this. The staging layer you flagged (the prior-reviews fetch, the rereview-mode CLI run, the scoped-diff swap, plus the existing provenance and new-scope steps that share the pattern) needs no model output, so it can all run as pre-workflow steps the agent just reads. Rather than special-casing this PR, I'll fold it into that migration so the whole layer moves at once and the eval producer runs the same pre-steps (otherwise the A/B stops measuring what production does). Merging this as-is since the dial defaults to `full` and follows the existing CLI-in-prompt pattern; the prompt mechanics here get deleted wholesale when the staging migration lands.

jwbron:
Follow-up for the staging migration discussed above (repo issues are disabled, so recording it here; the round-three plan carries it as step 5's second named slice):

# Stage a deterministic caller map

## The gap, observed live

The v1.5.0 behavior test (Khan/webapp#40793, seeded per the #40690 methodology) planted a cross-file amplification: a public function's `Limit(pageSize)` became conditional (`pageSize <= 0` disables paging, with a legitimate in-diff consumer on the prune path), while a pre-existing caller in an untouched file computes a page size that can reach the sentinel. The reviewer missed it entirely: no finding on the changed guard, nothing for the claim-validator to judge, nothing dropped at any gate (run 29284668567's artifacts confirm a pure finder miss). The near-miss makes the mechanism precise: first-principles named the exact hazard ("any future caller passing pageSize 0 silently gets an unbounded fetch") as a non-blocking design suggestion, but no agent ever enumerated the EXISTING callers, so "future hazard" was never upgraded to "present amplification" and the introduce-vs-amplify rule never engaged.

## Why this is structural

The pipeline deterministically computes facts about the diff (changed-line provenance, snap tables, routing, budgets) but nothing about the diff's dependents. "Who relies on the contract this change narrowed?" is left to model judgment under an investigation budget deliberately designed to discourage speculative exploration, and the camouflage is systematic: a contract change that ships with an in-diff consumer answers the cheap "why does this exist" question without any caller ever being read. Exhortation ("investigate harder") fights the budget discipline; the fix is a missing deterministic input, the same class of gap `provenance.json` and the anchor-snap table already closed for their questions.

## Proposed fix (a named slice of the deterministic-orchestrator staging migration)

1. **`callers.json` staging step (deterministic CLI in the lib).** For each exported symbol whose declaration or guard lines the diff touches, enumerate call sites across the repo (first cut: grep by symbol name filtered by package import; language servers later) and stage `symbol -> [file:line + snippet]`, capped per symbol, with a `truncated` marker when the cap bites. No model output needed, so it belongs in the pre-workflow staging layer (see the #246 review discussion).
2. **One rule in the removed-behavior audit.** A public function's guard, limit, or default becoming conditional is removed behavior for every existing caller: read the staged call sites before accepting an in-diff consumer as the justification, and classify introduce-vs-amplify per caller.
3. **Corpus case to price it.** A sanitized port of the #40793 D2 shape with a realistically reachable sentinel (small constant, not the original's edge-only arithmetic). It sits in the discriminating 20-80% band and gives the introduce-vs-amplify prose rule (#250) its first eval coverage. Recall-affecting change, so eval-gated with a targeted-repeats powered run per the memo's recipe.

Deliberately out of scope: an escalation edge that lets a design-altitude observation re-dispatch the correctness reviewer (model-driven control flow exactly where the roadmap removes it), and dynamic-dispatch/cross-service call resolution (real but much smaller residual than today's gap).

Refs: round-three plan step 5 (second named slice) and step 9 (first new mint); the behavior-test scorecard on webapp#40793; eval memo rev 5.1.

PR #247, opened 2026-07-09
review: dedupe lens discipline snippets into one staged disciplines file

jwbron:
## Summary

Fourth of five PRs in the fold-in batch two stack (stacked on #245; the stack is based on `main` since 2026-07-10). **The top of the batch-two stack is `jwbron/review-skill-rule-quote` (#248).**

The v1.4.0 re-run's cost premium vs the v1.3.1 baseline sits at dispatch time (uncached input +51-65%, cache writes +18-34% per run), partly because the discipline snippets (bounded investigation, untrusted input, lens-owned skills incl. quote-the-rule, the schema-rules trailer, the tri-state hunt contract) are stamped verbatim into every one of the eleven specialist-lens definitions and paid on every dispatch.

The mechanics, grounded in the gh-aw v0.81.6 source: `## agent:` sections are extracted from the markdown at compile time into per-agent definitions (`pkg/parser/sub_agent_extractor.go`), and each dispatch pays its definition's full text; the orchestrator's rendered prompt (the main body only) is written to disk at the engine-provided `$GH_AW_PROMPT` path. So the shared text now lives once, in a marker-delimited "REVIEW DISCIPLINES" section at the end of the main body, and:

- Step 1 stages it with one mechanical extraction: `sed -n '/^<!-- BEGIN ... -->$/,/^<!-- END ... -->$/p' "$GH_AW_PROMPT" > /tmp/gh-aw/review/disciplines.md`. The patterns are whole-line-anchored on purpose, so the sed command's own text and prose mentions of the markers can never open the range. A grep verifies the staged file carries the schema section; the fallback is a byte-for-byte heredoc copy, never a paraphrase. Prompt and disciplines version atomically (both live in review.md), with no coupling to the lib-checkout ref.
- Each lens definition now carries a pointer block ("Shared disciplines first: read /tmp/gh-aw/review/disciplines.md; its sections are part of this prompt") plus one "Domain notes" line preserving its own investigation examples. Its review rules, incident-derived hunts, and output JSON example are untouched.

Behavior-neutral by design: the instruction content is unchanged. Where the eleven stamped copies had already drifted (security-auth carried a fuller bounded-investigation and untrusted-input wording than the compact copies in the other ten), the disciplines section unifies on the fullest wording, so no lens loses instruction it previously had. The label-shape reviewers (correctness, skill-auditor, claim-validator, and the whole-change reviewers) are deliberately out of scope: their variants differ materially (per-agent investigation-cap CLI id semantics, `discussion` vs `evidence_trace`), so deduping them is not the mechanical win it is for the lenses.

Failure scenario this guards against, and how it is pinned: if the extraction silently produced an empty file, every lens would run without its disciplines (no quote-the-rule, no schema rules). `lib/disciplines.test.ts` replicates the exact sed range semantics against review.md and asserts the extraction is non-empty and carries all seven section headings, that every lens points at the staged file and carries no residual copy, and that the label-shape reviewers still carry theirs; Step 1 additionally verifies the staged file at run time before any lens is dispatched.

Net effect on review.md: 246 insertions, 474 deletions (about 16.6KB removed); each lens's dispatched definition shrinks by roughly 1.3KB. Whether that fully closes the measured dispatch premium is for the next lifecycle re-run to say; the duplication and the drift are gone either way.

Read-side arithmetic, stated plainly (raised in review by jeresig and the bot reviewer): each lens now reads the ~5.2KB staged disciplines file where its stamped copy was ~1.3KB, plus one tool round trip, and the disciplines section itself moves into the main body the orchestrator carries. So per-run tokens are at best a wash and may tick up; the change is justified by drift-elimination (a single source for what the lenses are told), not by a priced token saving. Under the deterministic-orchestrator staging migration (see #246 review discussion) this extraction becomes a pre-workflow step and the single-source file is the shape that work wants anyway.

cc @jeresig

## Testing

- `npx vitest run workflows/review`: 508 tests green (24 new in `lib/disciplines.test.ts`).
- `npx tsc --noEmit` clean.


jwbron:
Branch rebuilt on the refreshed chain (cherry-pick 813767c onto jwbron/review-out-of-lane per the #243 topology note; content unchanged apart from prettier formatting of the new test file). Reset local state to origin before continuing.

jeresig (approved):
I'm skeptical that this is going to have much impact - I think it's just moving stuff around? Isn't the caching cumulative based upon the previous input tokens? But I guess we can see in the testing.

jwbron:
You're right to be skeptical, and working the mechanics through makes it worse than a wash on the sub-agent side: each lens definition shrinks by ~1.3KB at dispatch, but each lens then reads the ~5.2KB staged disciplines file as a tool result (plus a round trip), and the section itself now lives in the main body the orchestrator carries. Caching discounts the repeats either way, so the dispatch premium this set out to close may not move; the body already defers that question to the next lifecycle re-run.

The reason to keep the change is drift, not tokens: the eleven stamped copies had already diverged (security-auth carried fuller bounded-investigation and untrusted-input wording than the other ten), and unifying on one source removes a correctness hazard in what the lenses are told. I've updated the description to state the read-side arithmetic plainly. Longer term this staging step migrates to a pre-workflow step with the rest of the staging layer (same direction as your #246 comments), where the single-source file is the shape that work wants anyway.

PR #248, opened 2026-07-09
review: render the quoted skill rule into the comment the author reads

jwbron:
## Summary

Fifth and last PR of the fold-in batch two stack (stacked on #247; the stack is based on #238, `jwbron/review-trial-skill`). **This branch, `jwbron/review-skill-rule-quote`, is the TOP of the batch-two stack**; downstream work (the re-review mode dial, the webapp preview runs) should base on it.

Quote-the-rule requires a lens skill finding to carry the exact rule text, but it lands in `evidence_trace`, which never reaches the PR; the author reads only `model_authored_prose`, a paraphrase of a rule they cannot check.

Failure scenario fixed: an author gets "**suggestion (non-blocking):** Datastore reads here should be strongly consistent per the datastore skill", has no idea what the skill actually says, either takes the bot's word for it or spends time hunting down the skill file, and cannot tell a faithful application from an overreach.

The change:

- `Finding` gains an optional `rule_quote` (the exact rule text, verbatim from the skill file). Optional, validated non-empty when present; `FINDING_SCHEMA_VERSION` stays 2 (no serialized finding is invalidated).
- The lens-owned-skills discipline (one copy, in the shared disciplines section after #247) tells each lens to fill it alongside the existing `evidence_trace` quotes, and the lens output JSON examples name the field.
- `renderComment` and the orchestrator's normalization step surface it into the posted comment as a `> **Rule:** <quote>` blockquote between the prose and any suggestion block. Only the wrapping is code-owned; the quote is skill-file text copied verbatim, per the determinism boundary.

The skill-auditor path is untouched: its prompt already requires both quotes inside `discussion`, which posts as-is.

cc @jeresig

## Testing

- `npx vitest run workflows/review`: 513 tests green (5 new: blockquote rendering and ordering, absent-field behavior, schema acceptance/rejection).
- `npx tsc --noEmit` clean.


jwbron:
Branch rebuilt on the refreshed chain (cherry-pick 7c07a42 onto the rebuilt jwbron/review-dispatch-tax; content unchanged apart from prettier formatting). This branch remains the TOP of the batch-two stack. Reset local state to origin before continuing.

workflows/review/review.md
```diff
@@ -3023,7 +3044,8 @@ Conventional-Comment `label` is emitted (the orchestrator computes it from
     "failure_scenario": "one sentence: the concrete inputs/state and the wrong outcome they produce",
     "producing_hunt": "hardcoded-user-facing-string",
     "model_authored_prose": "the comment the author will read",
-    "suggested_patch": "optional", "pre_merge_obligation": "optional"
+    "suggested_patch": "optional", "pre_merge_obligation": "optional",
+    "rule_quote": "optional: for a skill finding, the exact rule text, verbatim"
```
jeresig, line 3071: I assume most of these aren't coming from skills - do we need to include this for them?

workflows/review/review.md
```diff
@@ -1465,7 +1474,9 @@ entry; `failure_scenario` names the concrete failing scenario (specific
 inputs/state, then the wrong outcome) — it is the specific claim the
 claim-validator attacks, so make it checkable; `producing_hunt` names the hunt
 that produced the finding; `model_authored_prose` carries the entire human-read
-comment. Omit `suggested_patch`/`pre_merge_obligation` unless they apply.
+comment. Omit `suggested_patch`/`pre_merge_obligation` unless they apply; a skill
+finding also carries `rule_quote` (§Lens-owned skills), which the orchestrator
```
jeresig: §?

workflows/review/review.md
```diff
@@ -1439,7 +1444,11 @@ skill-level default or a per-rule `must`/`never`/`blocking` vs `should`/`advisor
 annotation) sets the finding's `severity`; when the skill declares none, judge by
 impact. Flag a skill violation only when you can quote **both** the exact rule
 text from the skill file **and** the exact violating line; put both quotes in
-`evidence_trace`, with no spirit-of-the-doc inference.
+`evidence_trace`, with no spirit-of-the-doc inference. Also copy the exact rule
+text, verbatim, into the finding's `rule_quote` field: evidence traces never reach
+the author, and `rule_quote` is what gets rendered into the comment they read (as
+a `> **Rule:** …` blockquote), so the author sees the actual rule, not a
```
jeresig: I don't think this cares about how it's rendered later...

jeresig (approved):
Great!

workflows/review/review.md
jwbron: Most aren't, right: the field is optional and a non-skill finding omits it entirely (the validator treats absence as fine). It's in each lens's example because any lens can own skills (the consumer's skills catalog maps skills to lenses by relevance criteria, so whichever lens owns the matched skill is the one that must emit rule_quote), and output examples are what the models actually copy; a field absent from the example tends to never get emitted. Cost is the one labeled line per definition.

workflows/review/review.md
jwbron: It does resolve (there's a "## Lens-owned skills" heading ~40 lines above, in the same disciplines text the lens reads), but the notation was needlessly cryptic; spelled it out as "the Lens-owned skills section above" in 57d6fbf.

workflows/review/review.md
jwbron: Agreed on the format detail: dropped the blockquote parenthetical in 57d6fbf. Kept the surrounding rationale (evidence traces never reach the author; rule_quote is what the author reads) since that's what tells the lens the field must be verbatim and author-facing rather than a paraphrase.

PR #250, opened 2026-07-09
review: state introduce-vs-amplify in the posted prose

jwbron:
Follow-up from the v1.4.0-preview amplification behavior test (Khan/webapp#40736, run [29051817727](https://github.com/Khan/webapp/actions/runs/29051817727)): the reviewer's behavior was correct (single blocking, validator-confirmed comment on the amplifying hunk; the merely-modified pre-existing defect stayed unposted), and the posted comment argued the provenance distinction in substance, naming the pre-existing default and attributing the regression to the guard removal. But it never said which part was which: the introduce-vs-amplify call lived in structure and implication, not in words the author reads.

## Changes

- The finder's pre-existing-bugs rule now requires the call as a plain clause in the `discussion` prose ('introduced by this change', 'pre-existing; this change amplifies it by removing the guard'), and covers the boundary case the test hit: the enabling mechanism predates the diff but the defect is new (a changed line drops the guard that made a pre-existing default safe), where the comment must name the mechanism as pre-existing and the regression as introduced.
- The claim-validator's amplification rule extends its `corrected` path symmetrically: add the missing clause whether the claim is an unlabeled amplification or an introduced defect riding a pre-existing mechanism.

Prompt-only; no code paths change. `pnpm run test --run` green (708), typecheck clean.

Stacked on #249.

### DIFF
--- a/.changeset/review-provenance-label.md
+++ b/.changeset/review-provenance-label.md
@@ -1,0 +1,5 @@
+---
+"review": patch
+---
+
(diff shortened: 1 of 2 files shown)


jwbron:
Eval-case gap worth closing here (or as a rider on #235): the live corpus has must-catch cases for the composite key, eventual consistency, and flag mock, but no amplification/provenance case. Khan/webapp#40736's seeded case (a 'redundant' WithLimit(5) removal amplifying a pre-existing Query default-limit-1 mechanism, plus a non-amplified pre-existing defect that must stay non-blocking, plus an off-diff observation that must not post) is fully specified in that PR's behavior-test record and passed live; porting it gives this PR's prose requirement a scoreable case and locks the gate behavior in regression.

jeresig (approved):

PR #251, opened 2026-07-09
review: re-review coverage for the eval corpus (open-PR cases + lifecycle chain)

jwbron:
Re-review coverage for the eval corpus. Stacked on #246 (the re-review mode dial); answers the gap that the corpus only modeled first reviews: every case was a fresh-PR snapshot, so re-review behavior (thread reconciliation, the flip decision, duplicate suppression, reduced-depth rosters) had no corpus instrument at all. The template is the measured seeded lifecycle (Khan/webapp#40730: planted bugs, a bad partial fix with added bugs, then the full fix), ported with the trial skill's sanitization gate (structural rewrite, generic naming, no webapp code or paths).

## Format: open-PR snapshots

A live-enabled case may carry `live.rereview`:

- `priorDiff`: the previously fully-reviewed diff. The staged prior review's hidden fingerprint stamp and the depth plan are derived from it with the same `lib/rereview-mode` functions production runs, so the dial's tripwire and depth decision are exercised, not simulated.
- `priorVerdict` / `priorDepth`: what the stamp records.
- `priorThreads`: the unresolved bot threads at this push, each with the opening comment (label template), an optional author reply, per-thread reconciliation ground truth (`expect: resolve|keep`), and optional mechanism alternates for duplicate detection.

Each lifecycle push is an independently runnable corpus case; a lifecycle is a chain of them (`golden-retention-lifecycle-1/2/3`). The prior state is authored ground truth rather than carried runner state, which keeps cases reproducible and lets the suite run any push in isolation; self-play lifecycles (the arm's own push-1 comments feeding push 2) remain the review-trial skill's job.

## What runs

- **Staging** (`live-stage.ts`): `threads.json` (with author replies, so the reconciler has the reply chain production gives it), `human-threads.json`, `prior-reviews.json` (stamped body), `rereview-plan.json`; under a reduced mode the scoped diff overwrites `full-stripped.diff`/`pr.diff`, exactly as review.md's scoped and flip-gated depths stage.
- **Producer** (`live-producer.ts`): dispatches `thread-reconciler` on rereview cases at every depth, and sizes the finder roster by the plan (`scoped` keeps the full roster, `flip-gated` keeps the correctness pass, `fast` reconciles only).
- **Scoring** (`rereview-match.ts`): thread-resolution accuracy against per-thread ground truth (unaccounted threads count as wrong, the accountability failure mode), the flip-gate input (kept blocking count, expected vs actual), and duplicate comments that re-raise kept threads. Fresh-defect recall on a push still rides the ordinary `mustCatchSpecs` through `live-match`.
- **A/B pricing** (`live-ab.ts`): `--re-review-mode <m>` sets the CANDIDATE arm's mode while the baseline stays `full`, so one run prices the dial on recall and dollars over the same cases; per-case and aggregate rereview metrics land in the JSON and markdown reports.
- **Deterministic replay** (`runner.ts` + `lib/verdict.ts`): `computeVerdict` gains `keptBlockingCount`, the flip rule as code: an otherwise-approving run is floored at REQUEST_CHANGES while a prior blocking thread stays open. Re-review cases replay the ground-truth reconciliation, render the accountability section into the planned body, and the suite asserts the push-2 case blocks even with zero posted blocking comments and the push-3 case flips to APPROVE with "All 4 prior review threads are resolved."

## The ported lifecycle

- `golden-retention-lifecycle-1`: four planted defects (deletion removes at most one row via a paging default, a cap off-by-one, an unbounded prune query, a test that asserts nothing).
- `golden-retention-lifecycle-2`: the bad partial fix. Two threads properly fixed (resolve), the off-by-one "fixed" into a different off-by-one (keep), the test fix deferred in a reply (keep), plus a fresh planted blocking defect in the new dedup (prefix-collision data loss) and an untested window. Expected verdict stays REQUEST_CHANGES.
- `golden-retention-lifecycle-3`: everything fixed; all four threads must resolve and the verdict must flip to APPROVE, with the intentional fire-and-forget prune error as a must-not-flag trap.

## Sweeping every dial setting

`eval/rereview-sweep.ts` prices all four modes in one command (the A/B's `--re-review-mode` compares one mode against a full-depth baseline; the sweep holds the prompt fixed and varies only the mode, which is the decision a repo faces when writing its ROUTING line):

```
pnpm dlx tsx workflows/review/eval/rereview-sweep.ts            # all four modes
pnpm dlx tsx workflows/review/eval/rereview-sweep.ts --modes scoped,fast
```

No case-format change is needed (the mode is a run parameter); three realities the report handles instead: every row shows the EXECUTED depth because the tripwire can override the dial (both lifecycle pushes diverge at share 0.67 and re-arm full under every mode, correctly); pricing the cheap paths therefore needs an under-threshold case, `golden-retention-fix-push` (a one-hunk fix push at share 0.2 whose fix resolves the prior blocking thread but plants a fresh defect inside the new hunk, so scoped/flip-gated must catch it); and `fast` has definitionally zero fresh-defect recall, shown as the mode's cost in the recall-vs-dollars table.

669 tests pass, lint and typecheck clean.


jeresig (approved):
Great! Thank you for adding tests for this.

PR #252, opened 2026-07-10
review: make the live eval a measurement tool (repeat aggregation, powered runs, drift watch)

jwbron:
Lands the tuning memo's rev-2 measurement items (memo: "Dialing in the evals", rev 2). The per-PR A/B stays a tripwire; this adds the machinery that turns repeats into resolution.

## What's here

**Repeat-aggregation report (`eval/aggregate.ts`, memo item 3).** Deterministic core + thin CLI: pools N `live-ab-report.json` artifacts (local paths or `gh` run ids) into per-case pass rates per arm with 95% Wilson intervals, pooled recall/verdict/noise rows, and true-miss vs found-but-dropped classification per gate bucket. Warns when a pool mixes review.md shas. Validated against the three 07-10 cumulative dispatches (29065968954 / 29065976210 / 29065977160): reproduces the memo's hand-computed numbers exactly (baseline 21/24 specs, 28/30 verdicts; candidate 22/24, 28/30).

**`--repeats <n>` in `live-ab.ts` (memo sizing section).** One dispatch runs the selected cases n times per arm (budget split per arm-run, fresh staging per repeat) and reports through the aggregation core instead of single-run percentages; this is the ~\$29 targeted-repeats powered run (`--cases <two anchor-fragile cases> --repeats 10`). The adversarial gate is decided by strict majority across repeats; the single-run best-of-three retry is skipped since the repeat structure buys the same flake evidence. The A/B workflow exposes `cases` / `repeats` / `force_arms` dispatch inputs.

**Scheduled drift watch (`review-eval-drift.yml`).** Weekly cron: full live corpus x3 repeats, both arms pinned to main's review.md (`--force-arms` past the identity short-circuit), aggregate published to job summary + artifact. Identical arms mean 6 arm-samples of the same prompt per run, so every drift run also refreshes the noise-floor bands, which the aggregate renders as data (memo item 6).

**incident-sql-missing-index diagnosed and fixed (memo item 2).** Read every recorded transcript in the wave + cumulative artifacts (28 arm-runs): the reviewer found the missing index in 28/28. The 8/16 catch rate was the spec accepting only the migration-file anchor while the reviewer anchored the same blocking finding at the hot query in `src/models/order.ts` (11 "misses") or had it provenance-dropped on an out-of-file mis-anchor (2, the adversarial-injection-approve pathology). Neither mechanism alternates nor lens routing were at fault; the case measured anchor-site preference. Specs gain `altLocations` (validated like the primary location; matcher, fallback filter, and drop classification all honor them) and the case accepts the query-site window. Replayed against all 28 recorded runs: 26 caught, 2 found-but-dropped, 0 true misses; the residual failure class is exactly the anchor-snap production defect, now correctly bucketed.

## Notes

- `pnpm test --run` (879 tests) and `pnpm typecheck` pass; eslint can't run in this worktree (dot-dir ignore), prettier applied.
- The aggregate CLI degrades per source: an unreadable artifact is listed and skipped, never fatal.
- Out of scope per the memo: the anchor-snap provenance-gate fallback (production reviewer change) and review.md semantics.


jwbron:
Post-review additions on this branch since the opening description:

- **Fallback match arbiter** (`eval/match-arbiter.ts`, pinned `claude-haiku-4-5-20251001`): fills the `MatchFallback` seam; unmatched specs only, same-file, capped 10/case, `via: "fallback"` audit trail, biased to refuse, degrades to non-match on API failure. On by default; `--no-match-arbiter` opts out.
- **Module split**: report shapes/renderers moved to `eval/live-ab-report.ts` (CI max-lines caught live-ab.ts at 1011; this worktree cannot run eslint locally).
- **Per-repeat artifact checkpoints**: a multi-repeat run now overwrites its out path after each repeat, so cancellation/crash cannot forfeit completed repeats (the always() artifact upload always has something to grab).
- **Measured noise floor published**: run 29069228968 (`--force-arms --repeats 3`, full 14-case corpus, $58.71 at the $60 clamp; 6 arm-samples of one prompt) measured recall 54-86%, verdict agreement 75-100%, noise 50-60%, judge quality 0.82-0.86. These bands render as data in every single-run report footer with provenance. The run also live-validated `--repeats` + the new dispatch inputs end to end.
- **Drift default budget raised to $85**: the corpus is 14 live cases on this stack and $60 trimmed case-runs (38/36 of 42).

Not landed, deliberately: baseline-arm reuse. Actions cache entries are unreadable across sibling/child PR runs, which is exactly the stacked topology carrying the 44% duplicate spend; the memo (rev 3) records the analysis and three candidate alternatives.

jwbron:
## Main vs stack tip: the cumulative measurement, and how to read it

Pooled from the three 2026-07-10 full-corpus dispatches (29065968954 / 29065976210 / 29065977160; baseline `origin/main` review.md `e54d8145`, candidate stack-top review.md `e403408c`, 3 repeats per arm, $42.83; table reproduced by `aggregate.ts <run-id> <run-id> <run-id>`):

| Metric | main | 95% CI | stack tip | 95% CI |
| --- | --- | --- | --- | --- |
| Must-catch recall | 21/24 (88%) | 69-96% | 22/24 (92%) | 74-98% |
| Verdict agreement | 28/30 (93%) | 79-98% | 28/30 (93%) | 79-98% |
| Noise (unmatched posted) | 24/45 (53%) | 39-67% | 23/45 (51%) | 37-65% |
| Misses (true / dropped) | 1 true / 2 provenance | | 0 true / 2 provenance | |
| Judge mean quality | 0.91 | | 0.90 | |

**Verdict: recall-neutral, which is the designed outcome.** Batch two is process and output-quality work (accountability, rendering, re-review machinery); the A/B's job for that class of change is to prove nothing silently regressed across a heavy prompt rework, and it did: the candidate arm has zero true misses. All movement sits inside the measured noise floor (recall wobbles 54-86% on byte-identical arms; see the report footer this PR adds).

**What the run actually taught us:**
- **The instrument was undercounting the reviewer.** Both weak cases were ruler error or gate pathology, not detection failure: the reviewer found the sql-index defect in 28/28 recorded runs (this PR fixes that spec via `altLocations`) and the injected auth defect in every adversarial run; what varied was anchor placement.
- **Anchor-snap is where the real gain is waiting.** Every observed drop (both arms, and production main itself) is right-file, right-mechanism, wrong-line, and each `golden-request-changes-authz` drop flips a verdict. Expected gain: +1-2 pooled specs and most of the verdict wobble, priceable at ~$29 with `cases=adversarial-injection-approve,golden-request-changes-authz repeats=10`.
- **The discrimination band has moved to the re-review cases.** On the 14-case corpus the identical-arm run pooled ~71% recall vs ~90% on these 10 cases: the #251 golden re-review lifecycle set is where recall-affecting work (mode dial, reconciliation) will show real movement next.

**Reading future reports:** the next drift runs will show recall and noise 'improving' because this PR fixed the ruler (altLocations) and de-noises matching (arbiter), not because the reviewer changed. Compare like with like via the review.md shas and the noise-floor provenance in each report.

jwbron:
One more commit (2369b86) from an adversarial pass over the instrument's own claims:

- **Ruler provenance**: every report now stamps its matcher configuration and a corpus content hash; the aggregate prints the ruler and **warns when a pool mixes rulers**. Without this, the drift series would show an unexplained step change whenever the instrument itself changes (arbiter default, corpus growth).
- **Honest noise-floor statistics**: bands now carry SD alongside min/max (min/max are extreme-value stats that only widen as samples accumulate), and identical-arm pools whose samples scored unequal case sets get a loud warning. The published v1 bands from run 29069228968 are so annotated: budget skips (38/36 of 42 case-runs) mean they fold case-mix variance in on top of wobble. Recall sd is 10 points; the 32-point min/max spread overstated pure wobble.
- **Rolling repeat budgets**: each arm-run's slice is now remaining-budget over remaining-arm-runs, so cheap early arms donate headroom forward; fixed slices were what caused the skip asymmetry above.
- **README statistical-honesty section**: clustered-interval optimism (pooled CIs are a lower bound on uncertainty), the repeats-gate being more permissive than single-run best-of-three (deliberate, but a relaxation reviewers should sign off on), and the arbiter's refuse bias being uncalibrated (audit `via: "fallback"` on close calls).

Known-open items live in the tuning memo rather than this stack: arbiter false-positive calibration, the gate-policy decision, a post-merge validation dispatch of the drift PR flow, and a clean re-measurement of the noise floor at the $85 budget.

jeresig (approved):

PR #262, opened 2026-07-14
review: always submit REQUEST_CHANGES with a non-empty body

jwbron:
GitHub rejects a REQUEST_CHANGES review event with an empty body, and Step 6 explicitly instructed submitting one ("...so submit it with an empty body", dating to #197). Under gh-aw v0.81.6 the safe-output flow posts the review event separately from the inline comments, so the comments do not make the event non-empty.

Observed live on the v1.5.0 behavior test (Khan/webapp#40793): the reviewer computed REQUEST_CHANGES correctly for two blocking findings, the submission failed with "Review body is required for REQUEST_CHANGES", and the PR was left with a COMMENTED review and no blocking verdict. The instruction survived this long because models usually filled the body anyway; this run followed it literally.

The fix: the REQUEST_CHANGES body's first line is now always `Changes requested — see inline comments.`, with the accountability section, note lines, and fingerprint stamp appended as before. The APPROVE empty-body case is unchanged (GitHub allows it).

cc @jeresig

### DIFF
--- a/.changeset/review-rc-body.md
+++ b/.changeset/review-rc-body.md
@@ -1,0 +1,5 @@
+---
+"review": patch
+---
+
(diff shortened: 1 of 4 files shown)


jeresig:
Ahh ok - that's too bad!

somewhatabstract:
:thumbsup:

PR #269, opened 2026-07-20
review: fix the corpus defects behind the drift run's stable misses

jwbron:
## What

Corpus fixes for three of the four stable misses in the 2026-07-20 drift report (#265). The triage (replaying the run 29724668102 artifact against the specs and the deterministic matcher) found that only one of the four was a genuine reviewer miss; the other three were defects in the corpus itself. This is a deliberate ruler change: catch rates will move because the ground truth moved, not the reviewer. The corpus sha stamp marks the break; see "Measurement" below.

### 1. Lifecycle trees: `void pruneNotes(...)` becomes `pruneNotes(...).catch(() => {})`

`golden-retention-lifecycle-3` expects a 0-comment APPROVE and documents the fire-and-forget prune as a deliberate must-not-flag trap, but the trap was mis-specified: `void` does not deliver the comment's own contract ("a failing prune must not fail the save"); it leaves an unhandled promise rejection. The reviewer's blocking flag on that line was technically correct and fired in every one of the 6 drift samples, forcing the wrong verdict on lifecycle-3 and injecting one guaranteed noise finding into every lifecycle-1 and lifecycle-2 run. The fix (applied to all three lifecycle trees, their diffs, and the embedded prior-push diffs, line counts preserved) makes the trap sound while keeping its intent: a documented deliberate pattern the reviewer must not flag. The trap spec gains a "swallow/empty catch" mechanism alternate so the must-not-flag guard still recognizes the new phrasing.

### 2. `retention-dedup-window-untested` spec: altLocation + mechanism alternates

Reported 3/6, wrong in both directions. All 3 recorded catches were Haiku-arbiter false accepts (each was the unrelated cap off-by-one finding at retention.ts:38, recorded `via: "fallback"`); meanwhile the reviewer surfaced the real dedup-coverage gap in 4 of 6 samples, anchored on `retention.test.ts`, which the spec's single path made ineligible for both deterministic and fallback matching. The spec now carries an altLocation on the test file plus two mechanism alternates covering the observed phrasing. Replayed deterministically over the recorded artifact: 0/6 as written, 4/6 fixed, with the prefix-collision spec still claiming its own finding first in all samples.

Both recorded findings are pinned as regression fixtures in `live-match.test.ts`: the real coverage finding must match deterministically, and the off-by-one finding the arbiter wrongly accepted must not match, guarding the widened mechanism against over-matching. The three recorded false accepts are also direct evidence that the arbiter's false-positive rate is uncalibrated; auditing its prompt is follow-up work, not attempted here.

### 3. `trial-dedup-eventual-consistency`: expected verdict APPROVE -> REQUEST_CHANGES

The case was internally inconsistent: it requires catching a defect that defeats dedup for exactly the retry/double-submit traffic the feature exists for, and both drift samples that scored it labeled the catch blocking, which mechanically forces REQUEST_CHANGES; the expected APPROVE was unreachable for a compliant reviewer. The recorded fixture moves to severity blocking at confidence 0.8, matching the trial-amplified-default-limit pattern, so the deterministic replay computes the expected verdict.

### Not fixed here: `retention-unbounded-prune` (0/6)

The one genuine reviewer miss. No sample ever flags the prune query loading the entire stale set (`pageSize: "all"`), and the reviewer's suggested fix for a different bug itself recommends the unbounded-read pattern. That is a review.md gap (no lens reasons about memory-bounded reads), left for a prompt change that the per-PR A/B can price.

## Measurement

Expected effects on the next drift run, all corpus-side: lifecycle-3 verdict agreement recoverable to in-band, dedup-window catch rate to roughly 4/6 with `via: "deterministic"`, noise down by one guaranteed finding per lifecycle-1/2 run, trial-dedup verdict agreement to in-band. This PR changes the corpus content hash, so the aggregate will (correctly) refuse to treat pre- and post-fix runs as one pool and warns on mixed rulers. Plan per the drift-report discussion: land this, then one manual drift dispatch re-baselines the noise-floor bands and starts the new series segment.

## Testing

- Full deterministic suite green (348 tests, including the two new matcher regression tests and the full-corpus replay gate that pinned the trial-dedup verdict).
- Matcher replay over the drift artifact ($0, no model calls): dedup-window spec 0/6 as written, 4/6 fixed; no cross-claiming with the prefix-collision spec.
- `pnpm typecheck` and prettier clean.


jwbron:
Closing in favor of the #270 stack: this PR and #270 are two triages of the same drift run that reached conflicting ground truths for the same files. Both fix lifecycle-3's `void pruneNotes(...)` and flip `trial-dedup-eventual-consistency` to REQUEST_CHANGES, but this branch fixes the `void` in all three lifecycle trees and keeps `retention-unbounded-prune`, while #270 keeps the `void` in lifecycle-1/2 as a planted must-catch defect (`retention-prune-unhandled-rejection`) and retires the unbounded-prune spec. The stack's treatment wins because #273 and #266 build on it and the Fable A/B (9 pooled dispatches, 30 cases x 3 repeats x 2 arms) was measured against the post-#270/#273 corpus; adopting this branch's trees would invalidate that measurement.

This PR's unique contributions were ported to #270 rather than lost:

- The `live-match.test.ts` regression fixtures pinning the drift run's two matching failures (the real test-file coverage finding must match deterministically; the arbiter's false-accepted off-by-one must never match).
- The dedup-window mechanism reconciliation: #270's tighter alternates replayed 3/6 deterministic vs this branch's 4/6; replaying the drift artifact showed the fourth sample is a genuine dedup-coverage finding phrased as "test does not exercise ... dedup", so #270 now carries this branch's `(test|coverage).{0,80}dedup` alternate (and not the bare `boundary` alternate, which was the arbiter's false-positive vector) and replays an honest 4/6.

The unbounded-prune spec this branch preserved is planned to return via #271 (the lens change that makes it a fair row); see that PR's measurement plan.

PR #271, opened 2026-07-20
review: teach the correctness and caching lenses to flag unbounded reads

jwbron:
## What

Closes the one genuine reviewer blind spot from the 2026-07-20 drift triage (#265, follow-up to #269/#270): `retention-unbounded-prune` was missed in all 6 drift samples, and the reviewer's own suggested fix for an adjacent bug recommended the same unbounded-read pattern (`pageSize: "all"`), so no lens was reasoning about memory-bounded reads at all.

Two prompt changes:

1. **Correctness lens, line scan**: unbounded reads and accumulation join the named defect classes; a query/fetch/scan that materializes a result set that grows with user data (missing LIMIT, `pageSize: "all"`, loading a whole table to act on a subset, unpaginated buffering). The rule also names the expected shape (page or batch it), so the fix pattern (a bounded read that deliberately processes one batch per invocation) is not itself flagged; that matters because `golden-retention-lifecycle-2/3`'s `PRUNE_BATCH` prune is exactly that shape and must stay clean.
2. **caching-resource specialist**: a matching "No unbounded reads" review rule and an `unbounded-read-materialization` tri-state hunt. The correctness lens is the always-on catch; the specialist adds depth when routed. (Note per review: no ROUTING routes `caching-resource` today, in this repo or webapp, so the correctness-lens rule is the operative change; the specialist half is the documented home for depth once a consuming repo routes storage-heavy paths to it.)

## Measurement

Recall-affecting by design. The original acceptance recipe targeted `golden-retention-lifecycle-1:retention-unbounded-prune`, but the corpus stack (#270/#273/#266) retires that spec: the lifecycle-1 facade documents no batch limit, so under a lens-less ruler the defect was ruled not evidenceable in-tree, and the slot now tracks the void-prune unhandled rejection. This PR is exactly the lens that makes the retired row fair again (the lens supplies the norm the tree does not state), so the plan is:

1. Land the corpus stack first, then rebase this PR onto main.
2. Re-add the retired `retention-unbounded-prune` spec to `golden-retention-lifecycle-1` in this PR (spec text is in git history at #270's parent), as the measurement instrument for the lens: without it, no corpus row tracks flagging an *undocumented* unbounded read (#270's review noted the gap).
3. Run the powered recipe:

```
gh workflow run review-eval-ab.yml --ref review-unbounded-read-lens \
  -f base_ref=origin/main -f cases=golden-retention-lifecycle-1 -f repeats=10 -f max_usd=30
```

Success: candidate arm catches `retention-unbounded-prune` at a real rate (baseline is 0/6 in the drift record) with lifecycle-2/3 clean on the bounded `PRUNE_BATCH` prune (no false flags), and no new flags on `mutation-retention-batch-delete-limit`'s already-caught defect beyond the existing catch.


jwbron:
Acceptance measurement complete, two iterations:

- **Iteration 1** (defect class appended to the line-scan enumeration, run 29763417032, $19.21): `retention-unbounded-prune` **0/10 vs 0/10**. The recorded runs showed why: the model posts one finding per defect site (the off-by-one claims the same query object) and treats "fetch all rows to delete them" as the job, not the bug.
- **Iteration 2** (current HEAD: mandatory bounds question per read call, two-findings-per-statement allowed, run 29767727038, $19.65): `retention-unbounded-prune` **9/10 caught vs 0/10 baseline** (Wilson 60-98% vs 0-28%). Verdicts 10/10 on both arms, other specs flat (one within-wobble miss on `retention-test-asserts-nothing`), unmatched-posted count unchanged, zero anchor snaps. Smoke A/B tripwire green.

Reports are on the two dispatch runs' artifacts (`live-ab-report`).

workflows/review/review.md
```diff
@@ -2857,6 +2867,9 @@ Skills index for this repo (read only the entries relevant to this lens's domain
   cache it feeds. `found` when invalidation is missing.
 - **`unbounded-cache-or-collection`** — a cache/collection with no eviction, TTL, or size
   bound. `found` when growth is unbounded.
+- **`unbounded-read-materialization`**: a read that loads an unbounded, user-data-sized
```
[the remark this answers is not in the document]
jwbron: Keeping the colon deliberately (house style here avoids adding em dashes in new text, and nothing parses these bullets). On the lower-confidence items: ROUTING is confirmed not to route caching-resource anywhere today (checked this repo and webapp), so the correctness-lens rule is the operative change; the PR body now says so, and the measurement plan is retargeted for the post-#270 corpus (that stack retires retention-unbounded-prune, and this PR is the lens that makes that row fair again, so the plan is to re-add the spec here after rebasing on the stack and run the powered recipe against lifecycle-1 with the lifecycle-2/3 bounded-prune negative controls).

workflows/review/review.md
```diff
@@ -1778,8 +1778,21 @@ Do two things in one pass over the files in the list:
    timing makes this line wrong? Look for logic errors (off-by-one, inverted
    conditions, null/undefined access, races, wrong-but-type-checking code);
    security issues (injection, XSS, unsafe deserialization, missing
-   authz/validation, SSRF, path traversal, committed secrets); and missing tests
-   for added/changed behavior (except pure docs or formatting).
+   authz/validation, SSRF, path traversal, committed secrets); and missing
+   tests for added/changed behavior (except pure docs or formatting).
+
+   Additionally, for **every query, fetch, or bulk-read call** the diff
```
jeresig, line 1784: This is incredibly backend specific... we'll want to split this out, for sure!

jeresig (approved):

workflows/review/review.md
jwbron: Added a note in my planning doc

PR #287, opened 2026-07-22
review: re-review fingerprint anchors on cache memory (body stamp never survives gh-aw ingest)

jwbron:
Plan-of-record: [The deterministic orchestrator](https://claude.ai/code/artifact/f988965d-1697-4e68-a9f2-d46a0915d225). Production bug surfaced by the two-arm live trial of the orchestrator stack (Khan/webapp#41006/#41007); independent of the stack, so it targets main directly. The stack merges this branch in so both trial arms exercise it.

## The bug

Every full-depth review renders a hidden HTML-comment fingerprint stamp into its review body (`renderRereviewStamp`), and the re-review plan CLI reads it back from `prior-reviews.json` to decide depth. That channel has never worked: gh-aw's safe-output ingest sanitizer strips ALL XML/HTML comments (`removeXmlComments` in gh-aw-actions `setup/js/sanitize_content_core.cjs`, a depth-tracking scan with no allowlist), so the stamp is deleted before the review posts.

Evidence, end to end, from the trial:
- The raw MCP call in a run's `safeoutputs.jsonl` contains the stamp; the ingested item in `safe-output-items.jsonl` does not (run 29893634730).
- None of Khan/webapp#40996's three production-posted reviews carries a stamp.
- 40996's re-review runs planned `"reasons": ["no-prior-fingerprint"]` and escalated to full depth (run 29872581602's `rereview-plan.json`), at 758 and 925 AI credits.

Consequence: the `re-review` ROUTING dial (`scoped`/`flip-gated`/`fast`), the flip gate, and the divergence tripwire are all silently inert in production; every repeat review pays the full roster. The dial was measured and shipped as the runs-per-PR cost lever; it has been buying nothing.

## The fix

The fingerprint gets a second carrier that demonstrably works: the Step 9 cache-memory record (`/tmp/gh-aw/cache-memory/pr-<n>.json`), which already stores `verdict`, the reviewed hunk signature, and `wasDraft` on every run.

- `stampFromCacheMemory` reconstructs a stamp from that record, validating every field (a gap returns null and the plan degrades to full, never cheaper). It prefers a new `stampHunks` field and falls back to `reviewedHunks`.
- `runRereviewPlanCli` reads body stamps first (unchanged priority; they cost nothing and become load-bearing again if the sanitizer ever allows them through), then the cache record, and records which carrier anchored the plan as `stampSource` in `rereview-plan.json` for observability.
- Step 9 (review.md) gains `stampHunks`: copied verbatim from `rereview-plan.json`'s code-computed value, never hand-computed. Step 1's added-lines-only hash is a different regime from the CLI's added-and-removed hash; mixing them would read as full divergence. `reviewedHunks` keeps serving comment scoping with Step 1's own hash on both sides.
- Hash-regime mismatches inside a stale cache record cannot be detected and surface as full divergence, i.e. a full review; cache eviction likewise degrades to full. Both are the pre-fix steady state.

## Verification

19 new vitest cases: `stampFromCacheMemory` field validation (verdict, wasDraft, hunk-map shapes, stampHunks preference and fallback), and the CLI fallback end to end (cache anchor reduces depth with `stampSource: "cache-memory"`, body stamp wins when present, no-anchor plans full, ready-for-review guard applies to a draft cache anchor, missing PR number and unparseable cache records degrade to full). Full review suite green (873 tests), typecheck clean, changed files lint clean.

### DIFF
--- a/.changeset/stamp-carrier-cache-memory.md
+++ b/.changeset/stamp-carrier-cache-memory.md
@@ -1,0 +1,7 @@
+---
+"review": patch
+---
+
+review: the re-review fingerprint anchors on cache memory; the body stamp never survives gh-aw ingest
+
+gh-aw's safe-output sanitizer strips all XML/HTML comments (`removeXmlComments`), so the hidden fingerprint stamp a review body carries never reaches the PR: every production re-review planned `no-prior-fingerprint` and silently escalated to full depth, making the `re-review` ROUTING dial (scoped/flip-gated/fast) inert. The plan CLI now falls back to the Step 9 cache-memory record (`verdict`, `stampHunks`/`reviewedHunks`, `wasDraft`) when no prior-review body carries a stamp, and records which carrier anchored the plan as `stampSource` in `rereview-plan.json`. Step 9 gains a `stampHunks` field copied verbatim from the plan CLI's own hash computation so hash regimes are never mixed. Cache eviction still degrades to a full review, never a cheaper one.
--- a/workflows/review/lib/rereview-mode.test.ts
+++ b/workflows/review/lib/rereview-mode.test.ts
@@ -13,6 +13,7 @@ import {
     runRereviewPlanCli,
     runRereviewStampCli,
     STAMP_SCHEMA_VERSION,
+    stampFromCacheMemory,
 } from "./rereview-mode";
 import type {HunkSignature, ReReviewStamp} from "./rereview-mode";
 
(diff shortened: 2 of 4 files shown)


jeresig (approved):

jwbron:
Superseded by #279, which absorbed this change wholesale as part of deterministic-orchestrator slice 0.

Verified file-by-file against `main`:

- `.changeset/stamp-carrier-cache-memory.md` and `lib/rereview-mode.test.ts` are identical.
- `lib/rereview-mode.ts` on main is a superset: same `stampSource` / `stampHunks` / cache-memory fallback logic, plus a refactor swapping the local `splitHunks` helper for the shared `splitPatchHunks` from `./diff`.
- `review.md` carries the same semantics, reworded for the new pre-agent staging: the sanitizer/cache-fallback note now lives in the `prior-reviews.json` description and the `stampHunks` cache-record block is intact. The agent-facing "stage them anyway" instruction is gone only because staging prior reviews is deterministic code now (`lib/stage-pr.ts`).

Rebasing onto main drops this commit as already-applied, leaving a zero-diff branch. Closing.

PR #294, opened 2026-07-24
review: move the whole reviewer roster to Opus 5

jwbron:
> **Stacked on #314** (AI-credits pricing overlay), which is where `claude-opus-5`'s rate is now defined. #315 (refusal visibility, formerly #311) is merged to `main`.

## What

Moves the **entire reviewer roster** to Opus 5 (`claude-opus-5`) — 22 model pins:

| Roles | Before | After |
| --- | --- | --- |
| orchestrator, `thread-reconciler`, `skill-auditor`, `claim-validator`, `conventions`, `documentation`, `holistic`, `completeness`, `test-adequacy`, 12 specialist lenses | `claude-opus-4-8` | `claude-opus-5` |
| `correctness-reviewer`, `first-principles` | `claude-fable-5` | `claude-opus-5` |
| `pattern-triage` | `claude-sonnet-4-6` | unchanged — deliberately the cheap first pass |

The eval's judge and match arbiter stay pinned to Haiku 4.5; they're the ruler, not the reviewer.

(22, not the 21 this PR originally described: `documentation` was added to `main` after these commits and moves with the roster, because the rebase applied the substitution to main's `review.md` rather than replaying an older copy.)

The basis: Opus 5 reports high precision **and** high recall at Opus 4.8's per-token price, which is half Fable's. Straight upgrade for the roles on 4.8; for the two on Fable it should keep the recall the 2026-07-20 A/B bought (82% → 89% must-catch, true misses down 41%) while dropping the +35% premium. **The powered A/B is the acceptance criterion, not this description.**

## Reading the A/B: four rows, four reverts

This is a roster-wide swap measured in aggregate, so a per-role regression is easier to miss than when one role moved at a time. Each of these reverts independently:

- **`correctness-reviewer`** — must-catch recall is the load-bearing metric. The six corpus rows Fable moved in #266 are where its gain lived: `incident-sql-missing-index:dm-default-backfill`, `trial-retention-deletion:erasure-default-limit-one` / `:erasure-ignores-delete-helper` / `:erasure-prune-error-swallowed`, `golden-retention-fix-push:quota-exceeded-never-true`, `golden-retention-lifecycle-2:retention-prune-unhandled-rejection`.
- **`claim-validator`** — the precision gate. Fable measurably failed here (noise 43% → 49%, one wrong blocking flag on a clean case), which is why it never left Opus. Opus 5's precision claim is the bet being made; read noise/unmatched-posted and any wrong blocking flag on the expected-clean cases.
- **`security-auth` lens** — see the refusal section below.
- **orchestrator** — a throughput row, not a quality one. Token-heaviest agent, 20-minute cap, and Opus 5 writes longer at the same effort. A timeout here is the swap surfacing operationally rather than measurably.

Also worth watching: whether Fable's noise regression (43% → 49%, attributed to its correctness reviewer posting more companion findings) comes back down, and whether `golden-retention-lifecycle-3`'s over-blocking on the expected-clean re-review push resolves.

## Refusals: measured, already open, and now mitigated

**This section previously described a risk this PR would take on. It is a hole that is already open, and #311 closes it.**

Eval run 30656579898 caught `correctness-reviewer` — on Fable 5, on shipped `review-v1.9.0` — refusing outright:

```
rawStopReason=refusal  tokens=2in/5207total
```

on `incident-auth-bypass` and `adversarial-injection-approve`. Not a context limit (5,207 tokens). Reproduced identically on two different harnesses, so it is the model.

The earlier reasoning here was that the specialist lenses were kept **off** Fable because its cyber classifiers can refuse benign security analysis. That reasoning was right and the mitigation went on the wrong roles: the 2026-07-20 A/B moved `correctness-reviewer`, a **default-roster** agent on the load-bearing metric, *onto* Fable. The shield went on the opt-in lenses; the default finder went the other way.

It was invisible because a refused agent emits no error. It emits nothing, and the review proceeds a dimension short.

**Opus 5 does not resolve this** — it ships elevated cyber safeguards of its own and can return `stop_reason: "refusal"`, so the hazard moves with the roster. What changes is that #311 mitigates it: a refused agent re-dispatches on `claude-opus-4-8` (already in that map for `claude-opus-5`), recorded as `fellBackTo`, one hop, never silent. With the roster on Opus 5, 4.8 leaves the roster entirely, which makes it a better fallback target rather than a worse one.

One correction to the earlier framing: *"gh-aw exposes no `fallbacks` parameter, so there's no server-side retry-on-another-model to lean on."* True of gh-aw, but scripted dispatch owns its runner, so the policy lives in our code instead.

**The detector changes character.** The drift signature named here — must-catch recall cratering on security-adjacent cases while everything else looks normal — is *masked* by #311, because a refusing reviewer now falls back and produces findings. In exchange the weekly live counters report the fallback rate per agent and model directly. That is a measurement rather than an inference, so it is a better detector, but it is a different one.

Refusals are also **intermittent**, not deterministic: probe run 30658862532 saw the same Fable pin clear both cases it had blocked. No single run confirms or clears a model.

`first-principles` still loses a property it was given on purpose: it ran Fable partly to be the one non-Opus reviewer, so its perspective diversity now rests entirely on its prompt. If its findings converge with `holistic`'s in the drift series, put one reviewer back on a different family.

## Pricing: now inherited from #314, and a hard merge gate

`claude-opus-5` is **not** in the api-proxy's curated AI-credits table at firewall v0.27.42 (it carries `claude-opus-4-5` through `4-8` and `claude-fable-5`, and stops). The guard rejects an un-priced model with a 400 *before* the request reaches the model, so every dispatch fails: the #266 failure, except now the whole roster is on the un-priced model rather than two opt-in agents.

This PR no longer carries its own `default-ai-credits-pricing` fallback. #314's `models.providers` overlay prices `claude-opus-5` at 50% of list, and it is strictly better than the fallback was: per-token rather than $/1M, it bills cache writes, and it is scoped per model instead of applying to *any* un-priced id (where a typo would silently bill at Opus rates).

**Consequence, and it is load-bearing: this PR must not merge before gh-aw v0.84.x is stable.** `apiProxy.providers` needs AWF v0.27.43; under v0.27.42 gh-aw drops the key silently, `claude-opus-5` is unpriced again, and every dispatch 400s. There is no fallback left to absorb that. #314 is inert but harmless under v0.27.42, so it can merge at any time; this PR cannot.

## Drift budget held at 220

Not re-derived from the list price. Per-token parity with Opus 4.8 isn't per-run parity: Opus 5 thinks by default and writes longer, and that now applies to *every* agent in a run rather than two. Under-sizing costs a red run (the `caseAsymmetry` gate) plus a contaminated noise-floor band — strictly worse than an unused ceiling. Re-size from a measured rate.

Note the budget is a **list-price** ceiling, as are all the eval's dollar figures: the firewall's pricing table and the SDK's `total_cost_usd` are both list, so real spend under our Anthropic discount is roughly half what the reports show. The guard therefore fails safe.

## Measurement

- [ ] Per-PR smoke live A/B (automatic) — tripwire only
- [ ] Powered A/B: full corpus x3, both arms — the load-bearing measurement
- [ ] **Cheap refusal probe first**: `-f cases=adversarial-injection-approve,incident-auth-bypass`, ~$1.57 and minutes. A refusing correctness reviewer would contaminate the recall column this PR is read on, so establish Opus 5's refusal behavior before buying the powered run.

Per the eval README, single-run deltas inside the noise bands are wobble, and judge quality/noise can move opposite to review health.

## Not touched

`.github/workflows/review.md` and `review.lock.yml` — this repo's installed reviewer, pinned to `review-v1.7.0`. Bumped at install time, guarded by `.github/workflows/review-pins.test.ts`.

## Checks

Deterministic tests pass; `tsc --noEmit` clean. `pnpm lint` fails locally on a duplicate-eslint-plugin clash between this worktree's `node_modules` and the parent checkout's — pre-existing, and this change touches no JS/TS.

## Build fixes folded in

Two problems that made CI red on this branch independently of the roster move, both in `package.json`:

- The `@anthropic-ai/claude-agent-sdk` range moved to `^0.3.219` without regenerating `pnpm-lock.yaml`, so `pnpm install --frozen-lockfile` failed with `ERR_PNPM_OUTDATED_LOCKFILE` before any job ran a test. The lockfile now matches.
- `zod` had been dropped from the root manifest. `main` carries it and `workflows/review/lib/dispatch-runner.ts` imports it, so removing it breaks resolution for root-level runs. Restored; nothing here explained the removal.

CI is green and the full suite passes (69 files, 1618 tests).


jeresig (approved):

jwbron:
still experimenting with this. Having some trouble with the gh aw firewall.

jwbron:
## The refusal risk this PR accepts is now measured, and one premise here is wrong

Context: #311. While benchmarking a different change I instrumented the eval's failure reporting, and it surfaced something directly relevant to this PR.

### `correctness-reviewer` is already refusing, on Fable 5, in production

[Run 30656579898](https://github.com/Khan/actions/actions/runs/30656579898), $1.57, two corpus cases:

```
rawStopReason=refusal  tokens=2in/5207total
```

Not a context limit (5,207 tokens). Not a harness artifact — reproduced identically on the Claude Agent SDK and on a second runtime. It is the model.

This PR's risk section reasons about refusals as a **specialist-lens** problem, on the grounds that the lenses were kept off Fable deliberately. But the 2026-07-20 A/B moved `correctness-reviewer` — a **default-roster** agent, on the metric this repo calls load-bearing — onto Fable 5, and it has been there since. The shield went on the opt-in roles; the default finder went the other way. What this PR describes as a risk it is *taking* is a hole that is already open, on shipped `review-v1.9.0`.

It was invisible because a refused agent emits no error. It emits nothing, and the review proceeds a dimension short. The eval reported it as `malformed output: output carries no parseable JSON object`, which sent ~$30 of runs after a contract bug that never existed.

### The premise to correct

> gh-aw exposes no `fallbacks` parameter, so there's no server-side retry-on-another-model to lean on.

True of gh-aw, and Anthropic's own refusal message points integrators at a fallback model. More to the point: **scripted dispatch owns its runner**, so the policy does not need gh-aw at all. #311 implements it in `lib/dispatch.ts` — on a refusal, the agent is re-dispatched on a fallback model, recorded as `fellBackTo`, one hop, never silent. `claude-opus-5` is already in that map, pointing at `claude-opus-4-8`, added pre-emptively for exactly this PR.

### What that means for the Opus 5 move

This PR flags that Opus 5 also ships elevated cyber safeguards and can return `stop_reason: "refusal"`, and calls the drift corpus the detector. Two changes:

1. **It is now a measurement, not a judgement.** Two named cases, ~$1.57, minutes: `-f cases=adversarial-injection-approve,incident-auth-bypass`. Worth running against the Opus 5 pins before the powered A/B, since a refusing correctness reviewer would contaminate the recall column this PR is read on.
2. **The nominated detector stops working, and is replaced by a better one.** The drift signature here is "must-catch recall craters on security-adjacent cases while everything else looks normal." With #311 merged that signature is *masked* — a refusing reviewer falls back and produces findings, so recall does not crater. In exchange the weekly live counters report the fallback rate per agent and model directly, which is a measurement rather than an inference. Net better, but the risk section should say the new thing.

Also worth folding in: refusals are **intermittent**, not deterministic. Probe run 30658862532 saw the same Fable pin clear both cases it had blocked. So no single run confirms or clears a model, and rate is the only meaningful reading.

### Suggestion

Stack this on #311 so the Opus 5 move lands with the hazard mitigated rather than accepted, and update the risk section for the two points above.

I attempted the rebase and stopped: this branch conflicts with `main` independently of #311 (the `sandbox.agent.version` retirement and the gh-aw v0.83.4 defaults have landed since these commits, and `review.md`'s frontmatter has moved underneath them). #311 does not touch `review.md` at all, so none of that conflict is mine to resolve — it needs whoever owns this PR's intent. Happy to do the rebase and the risk-section edit if you tell me how you want the sandbox-pin commentary reconciled.

jwbron:
## Rebased onto #311, and the pricing question resolved by checking

**Opus 5 is still not priced.** I checked `containers/api-proxy/ai-credits-pricing.js` at **v0.27.42** — the release gh-aw v0.83.4 defaults to now that this workflow's `sandbox.agent.version` pin is retired on main:

```
'claude-opus-4-5' … 'claude-opus-4-6' … 'claude-opus-4-7' … 'claude-opus-4-8'
'claude-fable-5'
```

That is the whole Anthropic section. No `claude-opus-5`. So this PR's `models.default-ai-credits-pricing` block is **still load-bearing**, and the blast radius is now worse than when it was written: main retired the sandbox pin on the grounds that v0.27.42 prices `claude-fable-5` — true, and irrelevant once the roster is on Opus 5. With all 22 pins on an un-priced model, a missing fallback 400s the orchestrator and every default-roster agent, not the two opt-in dispatches that #266 cost.

I dropped that block in the first pass of the rebase (I re-applied only the pin changes onto main's frontmatter). Restored in acf83a6 with the verification written into the comment. **The branch would have failed every run without it**, so this was worth the check rather than the guess.

One thing that did improve: the `gh-aw >= v0.83.0` floor this field needs is already met, since main compiles with v0.83.4. The compile-time constraint on consumers stands as documented, but this repo clears it.

## State of the rebase

Base is now `jwies/review-refusal-visibility` (#311) and the PR is **MERGEABLE** — it was CONFLICTING against main before, independently of the stack, because the sandbox-pin retirement and gh-aw v0.83.4 landed underneath these commits.

Resolution principle: keep main's infrastructure evolution, re-apply this PR's pin intent on top, rather than resurrecting frontmatter main has superseded. 22 pins now on `claude-opus-5` (the body says 21; main has gained one since). `pattern-triage` stays Sonnet 4.6; judge and arbiter stay Haiku. `version-sync` and `review-pins` guards pass, 1594 tests, `tsc` clean.

Two reconciliations I made in the README, both judgment calls worth a look:

- The refusal-fallback table still points **at** `claude-opus-4-8`. With the roster on Opus 5, 4.8 leaves the roster entirely, which makes it a better fallback target rather than a worse one: a genuinely different refusal profile instead of a neighbour.
- The refusal prose now says the hazard **moved with** the roster rather than being resolved by the pin change, which is what this PR's own assessment of Opus 5's cyber safeguards supports.

## Still owed: the risk section

The "Known risk this accepts" section still describes the pre-#311 world — refusals as a lens-only hazard, drift as the detector, and no fallback available. All three have changed, and the third was never quite right (gh-aw exposes no fallbacks parameter, but scripted dispatch owns its runner). That section is an argument in your voice, so I have left it for you rather than rewriting it; the material for the edit is in my earlier comment.

jwbron:
## Firewall pricing: v0.27.43 has Opus 5, but keep the fallback for now

Following up on the pricing check. **v0.27.43 (newest firewall release, and there is nothing after it) does carry a curated `claude-opus-5` entry**, at the rates this PR documents:

```
'claude-opus-5':  { input: 5.00, cachedInput: 0.50, cacheWrite: 6.25, output: 25.00 }
```

That entry is **strictly better than the `default-ai-credits-pricing` fallback**, because it bills cache writes — the one component the fallback path silently under-counts, which this PR's body already calls out as a known caveat.

Two reasons not to reach for it yet:

- **No released gh-aw defaults there.** gh-aw is at v0.84.1; a code search for `0.27.43` across the repo returns nothing. This workflow currently resolves v0.27.42, which does **not** price Opus 5 (verified: the table runs claude-opus-4-5 through 4-8, then claude-fable-5, and stops).
- **The only early route is a forward pin, which this file forbids.** The retired-pin note is explicit: *"Re-pin a version here only to hold a firewall release BACK, never to move one forward."* That is a convention rather than a technical limit, so it is overrulable — but doing so would freeze the firewall (giving up the per-container digests and future fixes) to buy cache-write billing on one model.

So the fallback stays. a08f433 records the removal condition in the frontmatter rather than leaving it to be re-derived: drop the `models:` block once a gh-aw release defaults the firewall to v0.27.43+, and do not pin forward to get there early.

## The `documentation` reviewer is included

It is the 22nd pin — added to main after this PR was written, which is why the body says 21:

- `review.md:2101` → `model: claude-opus-5`
- `README.md:539` → `| \`documentation\` | \`claude-opus-5\` | medium | Opt-in advisory targeted check |`

That is a side benefit of how the rebase was resolved: because the substitution was applied to **main's** `review.md` rather than replaying this PR's older copy, any role added upstream in the interim moves with the roster instead of being silently left behind on the old pin.

## Body edits still owed

Three, all in your voice so I have not made them:

1. **"21 model pins" is now 22**, and the role table is missing the `documentation` row.
2. **The tooling-unblock section** describes pinning `sandbox.agent.version` and needing gh-aw >= v0.83.0. Main has since retired that pin and moved to v0.83.4, so the floor is already met and the section's framing is stale — the field is still load-bearing, but for the reason above rather than the one written.
3. **The "Known risk this accepts" section** still describes the pre-#311 world: refusals as a lens-only hazard, drift as the detector, and no fallback available. Material for that rewrite is in my earlier comment.

jwbron:
Retargeted onto #315, which replaces #311 (accidentally auto-closed; same commits, new branch). This PR's own 3 commits and its `models:` block are unchanged from before that shuffle.

Separately: #314 adds a 50%-of-list `models.providers` overlay so AI credits meter at Khan's real Anthropic rate. It is independent of this PR now, but the two touch the same `models:` key, so whichever lands second needs to merge the blocks into one (a duplicate top-level key is silently resolved by YAML, and `gh aw compile` will not catch it because it only compiles `.github/workflows`).

jwbron:
Rebased onto #314 (AI-credits pricing overlay) now that #315 is on `main`. Description updated; the substantive change for anyone who already reviewed:

**The `default-ai-credits-pricing` fallback is now 2.5/12.5 $/1M, not 5.0/25.0.** #314's premise is that Khan bills Anthropic list minus 50%, so a credit should mean real spend. Leaving the fallback at list would have made the two pricing paths disagree depending on which one priced a given request. Flagging it because it is a semantic change to this PR's own block, not just a textual merge, and it is reversible if you would rather the fallback stay at list.

The two `models:` blocks are also merged into one key. Two top-level `models:` keys is a duplicate YAML key that resolves silently to whichever comes last, so one of the fallback or the overlay would have been dropped with no error; `gh aw compile` does not catch it because it only compiles `.github/workflows` and the duplicate was in the shared source.

Verified at this tip: frontmatter parses with one `models:` key, `gh aw compile` clean, full suite green (68 files, 1595 tests).

jwbron:
Removed `default-ai-credits-pricing` and the held-budget note, and rebased onto #314 (see updated description).

**One thing to be deliberate about before merging:** dropping the fallback means this PR now depends on #314's `models.providers` overlay to price `claude-opus-5`, and that overlay only reaches the firewall on AWF v0.27.43+. Under v0.27.42 (what gh-aw v0.83.4 defaults to, still `releases/latest`) gh-aw drops the `providers` key silently, so `claude-opus-5` is un-priced and the api-proxy 400s **every** dispatch, with nothing left to absorb it. That is the #266 failure across the whole roster rather than two opt-in agents.

So: #314 can merge whenever (it is inert but harmless under v0.27.42); this PR should wait for gh-aw v0.84.x to go stable. If you would rather not carry that constraint, restoring the fallback is a one-line revert.

jwbron:
Addressed the review feedback in 470b456 (push pending; the branch name is outside the sandbox push allowlist):

- review.md:182 (blocking, unpriced engine pin / no mechanical gate): two-part fix. The merge from the base branch had dropped the `default-ai-credits-pricing` fallback for the second time in this branch's history; restored, with the prose reconciled to the merged 50% providers overlay (fallback stays at list because in the only window it binds, v0.27.42 prices everything else at list too). New `model-pricing.test.ts` is the mechanical gate: every `model:` pin must have a providers overlay entry, and the fallback must exist while `claude-opus-5` is pinned. The external half of the gate is also satisfied: gh-aw v0.85.4 is now the latest stable (>= the v0.84.x floor).
- review.md:1178 (nitpick, stale comments): correctness-reviewer's Fable rationale replaced with the Opus 5 one; first-principles records the move; the overlay's "Current engine model" marker moved to `claude-opus-5`, with `claude-opus-4-8` relabeled as the refusal-fallback target.
- changeset slug: renamed to `review-roster-opus-5.md`.
- changeset:7 (question, why not carve out security-auth): the changeset now records the rationale: the hazard is intermittent and unproven on Opus 5, the one-hop fallback bounds each occurrence, and a uniform roster leaves one signature to watch and one revert to make.
- package.json:13 (lockfile sync): already resolved at head; `pnpm install --frozen-lockfile` resolves cleanly.
- changeset lens count: the flagged "twelve" text no longer exists at head; the PR description's "12 specialist lenses" is corrected to 11 (roster: 22 agents, 21 sub-agent pins plus the engine).
- review.md:1991 (suggestion, monoculture): deferred to measurement rather than resolved here. Planned: an A/B row pinning `documentation` to `claude-fable-5` against this branch's Opus 5 pin over the documentation live cases (including #331's readability cases; its clean case is the precision guard), `repeats` > 1. `first-principles`' disposition rides the same evidence.

jwbron:
Powered A/B complete (workflow_dispatch run 31738851830: full live corpus, repeats=2 per arm, baseline = this PR's base branch, so the delta isolates the roster swap from the pricing overlay). The stated acceptance criterion passes:

| Pooled metric | Baseline | Candidate (Opus 5 roster) |
| --- | --- | --- |
| Must-catch recall | 46/46 (100%) | 49/49 (100%) |
| Verdict agreement | 35/38 (92%) | 37/41 (90%) |
| Misses (true / dropped) | 0 / 0 | 0 / 0 |
| Cost | $39.18 | $40.13 |

Per the four revert rows:
- correctness-reviewer: the #266 Fable rows held on the candidate, including incident-sql-missing-index:dm-default-backfill (2/2) and golden-retention-fix-push:quota-exceeded-never-true (2/2); zero true misses anywhere.
- claim-validator: all three clean cases verdict-correct 2/2 on the candidate; no wrong blocking flag observed.
- security-auth: no refusals surfaced in either arm.
- orchestrator: no timeouts; candidate cost within 3% of baseline.

Adversarial hard gate PASSED on the candidate arm in every repeat. One row worth an eye at merge review: golden-documentation-stale-and-narrated (verdict) read 1/2 baseline vs 0/2 candidate, while both its specs were caught (the candidate labeled doc-stale-expiry-days blocking 2/2 vs baseline's severity split); a verdict disagreement with full spec recall on an advisory-documentation case is label calibration, not lost detection.

sxkosone (approved):

PR #298, opened 2026-07-28
autofix: opt-in, one-shot fixing of the PR reviewer's feedback

jwbron:
Adds an opt-in `autofix` workflow that addresses the PR reviewer's own feedback on demand.

Label a PR `autofix: blocking` or `autofix: nits` (they union), and the run fixes the reviewer's open threads in that scope, pushes one commit, replies in each thread, posts a summary, and removes the label. One run per arming.

## Scope of this first iteration

Deliberately narrow, per the design discussion:

- **One-shot only.** No continual cadence. The human re-arming the label is the loop, which for a trial is the better shape anyway: re-arm rate is the signal you want before automating it. Dropping cycles also removes the `workflow_run` chaining, the fail-closed cycle counter, the oscillation guard, and most of the governance surface.
- **Reviewer feedback only.** No human/author-sourced autofix.
- **No per-comment trigger.** See the design notes in the README for why an emoji trigger was dropped.

## The label axis model

The namespace is flat; the semantics are not. Three axes exist and the README documents all three, because `autofix: nits` and `autofix: loop` read as peers and are not:

| Axis | Values | Rule | Implemented |
| --- | --- | --- | --- |
| scope | `blocking`, `nits` | union | yes |
| cadence | `loop` (absent = once) | flag | no |
| source | `human`, `author` (absent = bot) | union | no |

A label on an unimplemented axis is **rejected, not ignored**. Honouring the blocking half of `blocking + loop` would present as a loop that mysteriously stopped after one cycle.

`isLoopEligible` encodes the one constraint that outlives v1: **nits never loop.** Non-blocking findings have no fixed point, so a nits-scoped loop cannot converge. It is in code rather than convention because it is the rule most likely to be violated by whoever adds the cadence axis.

## Division of labour

`lib/plan.ts` is the determinism boundary, mirroring the reviewer's split. Code resolves the scope, checks currency, builds the work list, and renders the trailer, all before the agent is asked to change anything. The plan is final: the prompt executes it or stops, never widens it or re-classifies a skipped thread. Nothing in `lib/` composes a sentence about the code under review.

Reuse rather than reimplementation: the label taxonomy comes from `render-comment.ts` (`BLOCKING_LABELS` / `NON_BLOCKING_LABELS`), the thread label parser from `rereview.ts` (`parseLeadingLabel`), and the currency fingerprint from `rereview-mode.ts`.

## Guards

Every refusal fails closed: no action, label removed, reason stated.

- **Review currency** is checked against the reviewer's hidden fingerprint stamp, **per file**. If the author pushed one unrelated fix after the review, findings in untouched files are still fixed and only the affected ones drop. An all-or-nothing gate would refuse routine PRs constantly.
- **Unparseable label → excluded.** Note this fails closed in the *opposite direction* from `rereview.ts`, where an unparseable label is treated as blocking so the thread is kept. Here the risk is an agent editing code on the strength of a finding it could not classify.
- **Outdated anchor, unreadable fingerprint (`hunks=overflow`), no review at all** → refuse.
- **Head moved mid-run** → abandon the push.
- **Never weaken a test** to satisfy a finding; the finding is left unfixed and reported. Prompt-level, not code-enforced; called out as such in the README.

## Why the bot token is load-bearing

`push-to-pull-request-branch` uses `KHAN_ACTIONS_BOT_TOKEN`, not `GITHUB_TOKEN`. GitHub creates no workflow runs for `GITHUB_TOKEN`-triggered events, so a push made with it would emit no `synchronize` and the reviewer would never re-review. That re-review is the **intended** verification for an autofix commit, which is also why autofix never resolves its own threads.

It is best-effort, not guaranteed: the chain has two links in a repo with the shared push-triggered reviewer and four in Khan/webapp, and the live trial lost one to a gh-aw setup failure that was invisible on the PR. The token is load-bearing for the stronger reason that `GITHUB_TOKEN` guarantees zero re-review while the bot token buys a best-effort one. The summary comment states the pending status on every push, and the human re-arming loop is the accepted backstop for v1; the README's "Verification is best-effort" section has the details and the numbers.

## The commit trailer

Every autofix commit carries `Autofix-Version` / `Autofix-Scope` / `Autofix-Cycle` / `Autofix-Threads`. v1 never reads it back. It is written because the branch is the only cycle store that survives cache eviction and needs no external state (the same reasoning that put the reviewer's fingerprint in the review body rather than cache memory), and because diffing `Autofix-Threads` against what the next review still reports open is how the trial scores whether a fix actually cleared a finding.

## Security review

`gh aw compile` flagged two new restricted secrets on this workflow. Both are already in use by `review.lock.yml` in this repo:

- `ANTHROPIC_API_KEY` — the `claude` engine, identical to the reviewer's use.
- `KHAN_ACTIONS_BOT_TOKEN` — the push, for the reason above. Already used by the reviewer for `resolve-pull-request-review-thread` and `add-reviewer`.

Two further secrets appear in the compiled manifest that I did not add; both are gh-aw v0.83.4 boilerplate and are unset in this repo, so they compile to empty:

- `COPILOT_GITHUB_TOKEN` — read only by gh-aw's "Check for OAuth tokens" preflight step.
- `GH_AW_CI_TRIGGER_TOKEN` — gh-aw's optional CI-retrigger token on the safe-outputs job.

No new actions beyond gh-aw's own pinned set. The agent job is `contents: read` / `pull-requests: read`; all writes go through safe outputs. `roles` requires write access (deliberately **not** the reviewer's `roles: all` override), forks are excluded, and `skip-ai-review` PRs never start.

## Things worth a second opinion

- **Bootstrapping:** `autofix.md` pins `autofix-v0.0.0`, which does not exist as a tag yet. The version-sync script rewrites it to `autofix-v0.1.0` in the Version Packages commit, so the workflow is only runnable after the first release. Same shape as any new workflow package here, but worth knowing before trying to trigger it.
- **Labels need creating** (`autofix: blocking`, `autofix: nits`) in whichever repo trials this first.
- The **bot-fixes-bot-then-bot-approves** concern from the design discussion is much softer with one-shot plus human re-arming, so I have not added a "requires human approval" gate. Worth revisiting before the cadence axis lands.

## Testing

64 unit tests across the five lib modules. Also smoke-tested the CLI end to end under `tsx` exactly as the prompt invokes it, against staged fixtures, confirming the armed, stale-path, and no-review paths.

Full CI locally: lint, test, typecheck, build all clean.


workflows/autofix/autofix.md
```diff
@@ -0,0 +1,598 @@
+---
+description: >
+  Addresses the PR reviewer's own feedback on demand. Opt in per PR with an
+  `autofix: blocking` or `autofix: nits` label; the run fixes the reviewer's
+  open threads in that scope, pushes one commit, replies in each thread, and
+  removes the label. One run per arming.
```
jeresig: It seems like it would be better (if this is only to trigger a single run) to instead use the `/fix` or `/autofix` convention.

(Although, I'll admit that I'm not sure what value this is providing right now since I would imagine that if folks wanted this they could just have their harness handle this process for them.)

workflows/autofix/autofix.md
```diff
@@ -0,0 +1,598 @@
...
+# so a comment from someone without write access never reaches the agent.
+#
+# The command match is written out longhand rather than using gh-aw's
+# `slash_command` trigger. gh-aw's compiled gate only matches the command
+# followed by a space, a bare `\n`, or end-of-body, so a comment saved with a
+# trailing CRLF — which the GitHub web UI produces when you press Enter after
+# the command — never activates the workflow. That silently killed `/review` in
+# Khan/webapp#40943. `scope.ts`'s parser tolerates the same shapes; keep the two
+# in step.
+if: "(github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name == github.repository && startsWith(github.event.label.name, 'autofix: ') && !contains(github.event.pull_request.labels.*.name, 'skip-ai-review')) || (github.event_name == 'issue_comment' && github.event.issue.pull_request != null && (github.event.comment.body == '/autofix' || startsWith(github.event.comment.body, '/autofix ') || startsWith(github.event.comment.body, '/autofix\n') || startsWith(github.event.comment.body, '/autofix\r') || startsWith(github.event.comment.body, '/autofix\t')))"
```
jeresig: It seems like this is handling `/autofix` comments but nothing is triggering when there is a comment like that left - I think this may need to be re-worked?

jeresig (approved):

workflows/autofix/autofix.md
jwbron: Both the `/autofix` and label convention are supported.

Agreed that as-is this isn't particularly interesting at the moment, it's more of a base to build on. I imagine using this when I'm not at my machine, allowing me to address bot feedback from my phone. The useful bit comes next. I'm going to add:
- support for preemptively adding this autofix label, allowing the autofixer to immediately respond to bot feedback.
- per-user configuration, allowing each user to have a global preference.
- support for re-review cycles

I think the above will be a nice improvement to workflows. If the blocking bot feedback is consistently sound, we should generally be able to have bots automatically address it. Nits/questions are tougher, but if we can at least have a baseline "correctness assurance cycle" to harden PRs, we could reduce load on engineers and reviewers. Fewer bot comments to have to read and fewer issues in code for reviewers to deal with.

workflows/autofix/autofix.md
```diff
@@ -0,0 +1,610 @@
...
+#   1. Same-repo branches only. A fork PR gets no secrets, so the push would
+#      fail anyway.
+#   2. The label that fired this event is an autofix label. Every other label
+#      addition on the PR is a run we never pay for.
+#   3. Never on a PR the reviewer was told to skip: with no review there is
+#      nothing to fix, and the plan would refuse in Step 2 regardless.
+#
+# COMMAND PATH — deliberately weaker, and worth understanding before you touch
+# it. `issue_comment` carries no `github.event.pull_request`, so the fork guard
+# and the `skip-ai-review` check CANNOT be evaluated here at all; they move into
```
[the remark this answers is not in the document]
jwbron: Fixed in 9037a9b (pushed shortly after this review landed, so it is not reflected above).

Both refusals are now real code in `buildPlan`, ahead of any work planning:

- Staging records the head's fork status in `context.json`; `plan.ts` refuses on `input.isFork !== false`, so unknown or unreadable fork status refuses rather than proceeding. A staging gap cannot reopen the hole.
- `plan.ts` refuses when `input.labels` includes `skip-ai-review`.

Both are enforced on every path, not just the command one: a duplicated guard is cheap and a missing one authorises a code push. `plan.test.ts` covers both refusals plus the unknown-fork case.

The prose was corrected in the same commit rather than left as the aspiration this finding caught: the `autofix.md` comment and README now say the checks are enforced in `plan.ts`, note that the command path still burns a job before refusing, and record why the checks cannot live in the `issue_comment` branch of the `if:` at all.

Same commit also threads `AUTOFIX_BOT_LOGIN` through to the worklist's ownership guard, which is the non-blocking `plan.ts` suggestion in this review.

workflows/autofix/lib/stage.ts
```diff
@@ -0,0 +1,456 @@
...
+        graphql: async (query, variables) => {
+            const res = await fetch(`${api}/graphql`, {
+                method: "POST",
+                headers: {...headers, "content-type": "application/json"},
+                body: JSON.stringify({query, variables}),
+            });
+            if (!res.ok) {
+                throw new Error(`GraphQL failed: ${res.status}`);
+            }
+            return res.json();
```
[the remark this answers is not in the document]
jwbron: Fixed in 7816541. Agreed on the direction and on the diagnosis; the chain you traced (200-with-`errors` → no connection → `break` → empty `threads.json` → no-op plan → `labelsToRemove` populated → label cleared with findings open) is exactly right, and the REST paths' own `throw` is the right precedent.

Taken as suggested, with two deliberate choices:

- **Any `errors` entry is fatal, including alongside partial data.** Your sketch guards a body carrying `errors` regardless of `data`, and that is the behaviour implemented. Staging a subset would be worse than refusing: the plan would fix whichever threads arrived, clear the label, and report a clean run over the ones that did not. There is a test for the partial case specifically (`onePage` content plus an `errors` array).
- **Guarded at both the transport and the reader**, rather than only the port. Duplicated on purpose, the same reasoning as the fork guard in `plan.ts`: the check is cheap and its absence clears an arming label on a PR with live findings. It also matters that the reader is the layer the unit tests can reach, since the real port is constructed inside the `require.main === module` block and is not importable.

`collectThreads` now throws rather than `break`s when a page yields no connection, unconditionally rather than only on the first page: a later page missing its connection is equally malformed, and the loop already terminates normally on `hasNextPage !== true`.

One thing your finding did not mention that fell out of fixing it: the test harness defaulted unsupplied GraphQL pages to `{}`, which the old code read as "no threads". That default was itself the fail-open shape, so it now returns a well-formed empty page and the error paths are exercised only by the tests that mean to.

Four tests added on `collectThreads` (rate limit, partial data, malformed body, and a well-formed empty PR that must still return `[]`) plus four on the extracted `assertNoGraphqlErrors`. I checked the guards are load-bearing rather than vacuous by reverting them: 3 of the 4 new `collectThreads` tests fail, 27 pass, and the well-formed-empty case correctly passes either way. Full suite 1545 pass, lint/typecheck/build clean.

No lock recompile: the change is in `lib/`, which is checked out at runtime from the pinned ref, not embedded in `autofix.lock.yml`.

jwbron:
## Follow-ups from the verification-gap investigation

Recorded here rather than as issues, because issues are disabled on both this repo and Khan/webapp. Neither is in scope for this PR; 7aed595 covers only the prose and the one line of output.

### 1. Khan/webapp's reviewer: `status-comment: false` → `true`

One line, in the consumer, and it fixes strictly more than anything autofix-specific: every failed reviewer run would say so on the PR, autofix pushes and human pushes alike. Today none of them do, because webapp's reviewer is an `issue_comment` local override (so its runs never join the PR's check suite), `status-comment` is false (so it posts nothing), and issues are disabled (so gh-aw's `failure-report-as-issue` fallback 410s). Of that repo's last 100 reviewer runs, 15 of the 53 that started ended in `failure`, and none of those 15 was visible on the PR it was reviewing.

The cost is exactly what the override deliberately suppresses: a run-status comment per review. That is the reviewer's and webapp's call, not this PR's.

**Unverified:** whether gh-aw's status comment updates in place or posts twice. It compiles out when the flag is false, so there is nothing in the lock file to read. Confirm before adopting.

### 2. The deferred re-review detector

Not blocked, and it needs nothing added to autofix's trailer. The question it asks is: *does a review by the reviewer bot exist whose `commit_id` is this autofix commit or later?* That is answerable from the commit SHA alone.

What it needs is a **home that runs later than the autofix run**. Autofix is one-shot and synchronous with its own push, so at the moment it posts its summary the re-review has not been attempted yet; it structurally cannot observe the outcome. That makes any detector a new scheduled sweep or a `workflow_run` watcher, which is where the cost is.

The `issue_comment` variant is the harder one: a `workflow_run` watcher has to map the run back to a PR, and `run.pull_requests` is empty for repository-level events.

Worth building when autofix gains the cadence axis and stops having a human in the loop. Until then the human re-arming loop is the backstop, and the point of 7aed595 is that the summary comment now tells that human they are it.

PR #299, opened 2026-07-29
review: add an opt-in `documentation` reviewer

jwbron:
Adds an opt-in `documentation` reviewer: advisory-only, whole-change, checking the **comments and prose docs a diff adds or changes** against a documentation policy. Off everywhere until a repo adds `enable documentation` to its `ROUTING` file. First of three PRs from the #ai-guild thread on LLM-generated comments; the premise is that comment quality is an enforcement problem, not a prompt problem.

**Flags:** a comment that restates the code; one that narrates the change ("updated to…", "previously we…"); one this diff falsified; commented-out code; and the other direction, a *specific* unexplained constant, workaround, or ordering constraint the change introduces.

**Does not flag:** density preferences, anything CI owns, generated files, a `TODO` with a ticket, or comments the diff did not touch.

## Design notes

**The policy lives inline in the agent definition**, like the specialist lenses' `Review rules`. It has to be written down somewhere, because every finder must quote the rule and `claim-validator` drops what it cannot verify against real rule text; an unwritten policy just yields `{"findings": []}`. A fifth required `.github/aw/review/*.md` consumer import was the alternative, rejected because the baseline is universal and a missing runtime import fails the whole review at run time, so every repo would have to write the file before its next PR could be reviewed. Repo-specific calibration rides the per-directory `REVIEW.md` contracts that already exist, which also answers Jeff's "backend and frontend may want different outcomes": they can, per directory.

**It never reasons about who wrote the text.** It cannot tell human from model, must not guess, and the policy is the same either way. Attribution is unreliable anyway (we strip Claude trailers by convention, so absence proves nothing), and a rule binding only bot-authored PRs is bypassed by deleting a trailer. If we want the author axis, it belongs in the counters as a measurement, not in the rubric.

**The label is load-bearing.** Findings render as `suggestion (non-blocking, documentation)`, code-assigned from the lens like the `, best-practice` variants. The stacked autofix PR depends on it: `worklist.ts` selects threads by parsing the label off each **posted comment**, so nothing else about a finding survives to reach it, and without the variant a documentation-scoped autofix would widen to every nit on the PR. There is deliberately no blocking variant; minting one would enlarge `BLOCKING_LABELS`, the set `autofix: blocking` acts on. Blast radius is small because the taxonomy is centralized (`thumbs-sweep-github.ts` spreads both sets, `verdict.ts` counts blocking only, the eval runner goes through `labelForFinding`). The README notes the consumer ordering: autofix reads labels minted by whichever reviewer version a repo has *installed*.

**Anchoring.** The best finding here is "the change falsified this comment", and that comment usually sits on an *unchanged* line, because what changed is the code below it; the provenance gate would drop it. So the prompt says to anchor on the changed code line and name the comment in the prose. The golden case pins it: the stale-comment finding anchors on `EXPIRY_DAYS = 90`, not on the comment above.

## Supporting changes

- **`claim-validator` gains a documentation-claim rule.** A documentation claim is neither a code claim nor a `skill` claim, so it was being validated as a runtime-behaviour claim. The new bullet says what to verify and names the characteristic false positive: mistaking a real constraint for a restatement.
- **The live eval producer now dispatches a case's `enable`d reviewers.** `produceLive`'s roster was `DEFAULT_FINDERS + routing.lensesToSpawn`, so **no opt-in whole-change reviewer has ever had a live arm** and none could earn its `enable` line the way the README requires. `routerConfig.enabledReviewers` now adds them in canonical `ENABLEABLE_REVIEWERS` order, an unrecognised name throws (a typo would otherwise produce a green, expensive run measuring nothing), and `labelLens` gains all six. Existing cases set nothing, so the corpus is unchanged.
- **Reconciliation with scripted dispatch** (third commit, from the rebase onto #289). Main moved the shed ranking and label taxonomy out of `review.md` prose into code, so those hunks moved with them. One gap the textual merge could not see: the dispatcher code-assigns each label-shape reviewer's `lens`, hardcoded `correctness` for everything but `skill-auditor`, so a documentation finding reached the render stage with the wrong lens. Replaced with a named `LABEL_SHAPE_LENS` map, with a test.

## Corpus

A matched pair, both live-enabled:

- **`golden-documentation-stale-and-narrated`**: a sound behaviour change (expiry 30 → 90 days, clock injected) leaving a falsified comment and a change-narrating comment. Both must post, verdict stays APPROVE. Plus a must-not-flag trap, an `EXPIRY_SLACK_MS` comment that reads like a restatement but records a constraint the code cannot show; it is inside the diff, so a false flag on it is reachable rather than gate-dropped for unrelated reasons.
- **`clean-documentation-earned-comments`**: the precision guard, and the more important of the two. Every added comment earns its line, the run must post nothing, and both traps sit on added lines next to the code they explain.

Both diffs verified against `computeChangedLines` (zero orphan hunk lines); annotated line numbers match the fixture trees.

## Left for later

- **PR titles and descriptions** (Mike's original complaint). The read side is nearly free; the write side needs gh-aw's `update-issue` safe output and a shrink-only guard. Separate PR, not stacked.
- **Enabling it anywhere.** No `ROUTING` change ships here; it earns its `enable` line through the live A/B, which the producer change above finally makes possible.

## Testing

1421 tests pass, typecheck clean. The corpus pair runs through the deterministic replay in `suite.test.ts`, where `checkExpectation` enforces the `mustCatch` ids and the pinned `postedCommentCount`.

## Worth a second opinion

- **`documentation` is first-to-shed**, ahead of `conventions`. Advisory comment findings should go first under budget pressure, but the interior order is editorial.
- **One label, not four**, where `conventions` picks among four: exact autofix selection and nil verdict impact, at the cost of the nitpick/question/note nuance.
- **`claude-opus-4-8` / medium** is copied from `conventions`, not measured.


workflows/review/review.md
```diff
@@ -2685,6 +2696,137 @@ concrete cost of the deviation if it stays (a convention with no statable cost i
 not worth flagging). If nothing deviates from repo conventions, return
 {"findings": []}.
 
+## agent: `documentation`
+---
+name: documentation
+description: Advisory, opt-in check that code comments and prose docs in the diff document intent rather than restate code; returns findings as JSON.
+model: claude-opus-4-8
```
jeresig, line 2048: 4-8? not 5?

jeresig (approved):

workflows/review/review.md
jwbron: I'm having some trouble getting opus 5 working. Going to get that sorted out in #294

PR #302, opened 2026-07-30
[🔥AUDIT🔥] review: fix open-thread suppression, unreachable on every conforming run

jwbron:
🖍 _This is an audit!_ 🖍

## Summary:
`openThreadsFromStaged` accepted an opener author of `github-actions[bot]`
only. `get_review_comments`, the tool review.md tells the orchestrator to
copy `threads.json` from verbatim, renders that same account as bare
`github-actions`, and the filter fails closed — so a *correct* staging
yielded zero usable threads and suppression never ran. It also read `path`
from the thread while the tool carries `path` per comment, and
`suppressOpenThreadDuplicates` matches on `path`, so a thread that cleared
the author check still suppressed nothing.

Measured on webapp#41197's three-round seeded lifecycle (the review-v1.8.0
acceptance trial): `threadSuppressions: []` in all three rounds while the
re-reviews duplicated open threads. 6 of round 3's 8 comments landed on the
exact path and line of an already-open thread, and the reconciler's `keep`
list held those thread IDs in the same run, so the data was present and
unusable.

Every unit fixture spelled the bot the way the code did, which is why the
suite passed throughout. The new cases use the tool's real shape verbatim,
assert an end-to-end suppression (with the blocking thread still flooring
the verdict) rather than only the parse, and keep a human-opened thread
refused so the widened author check cannot drop a finding outright. Both
new shape cases fail against the old filter, verified by reverting it.

Second half, since a fail-open guard that cannot be seen failing is how
this survived a release: `stagedThreadShapeFailure` reports a staging whose
shape defeats the filter, as `threadSuppressionUnavailable` on
dispatch-result.json plus a run-log warning. "Nothing to suppress" is no
longer indistinguishable from "suppression silently did nothing".

The condition lives in dedup.ts beside the filter it describes, which also
keeps dispatch.ts inside its 1000-line lint ceiling.

## Test plan

- `vitest run workflows/review/lib/dedup.test.ts`: 23 pass (20 before; 3 new).
- **Guards proven load-bearing:** reverting the author and path fixes fails both new shape cases and leaves the human-opened-thread guard passing, which is the correct split.
- `vitest run workflows/review/lib/dispatch.test.ts`: 30 pass.
- `tsc --noEmit` clean; `eslint` clean on all three changed files (`dispatch.ts` lands at exactly its 1000-line ceiling, which is why the new condition lives in `dedup.ts`).
- Pre-existing on `main`, unrelated and untouched here: `lib/dispatch-runner.test.ts`, `eval/live-ab.test.ts`, and `eval/rereview-sweep.test.ts` fail identically on a clean `origin/main` worktree (7 tests, same set).

## Follow-up

The deeper issue this exposes is that `threads.json` is still prompt-staged: `stage-pr.ts` defers it ("Phase 2; a later slice"), so slice 3 moved the *decision* into code while leaving its *input* model-produced. This PR makes the resulting fail-open visible rather than closing that seam. Staging threads in code is the real fix and belongs with the remaining slice work.


somewhatabstract:
:thumbsup:

PR #304, opened 2026-07-30
[🔥AUDIT🔥] review: split the roster concern out of dispatch.ts, which is over its line cap

jwbron:
🖍 _This is an audit!_ 🖍

## Summary:
main is red on lint, and neither PR that caused it could have seen it.
`@khanacademy/eslint-config` sets max-lines to 1000. #302 took
workflows/review/lib/dispatch.ts to exactly 1000 lines; the single line #299
added to SHED_RANKING took it to 1001. Each PR was green against its own
base, so the violation existed only in the merge, which is invisible to a
per-PR lint run. Every open PR in the repo inherits the failure, including
#300.

Fixed by extracting a concern rather than by reclaiming a line, so the next
addition does not land in the same place, and rather than raising the cap,
which lives in the shared Khan config and would be a house-rule deviation.
DEFAULT_FINDERS, SHED_RANKING, Roster, RosterShed and computeRoster move to
dispatch-roster.ts and are re-exported from dispatch.ts, which already
advertises one import surface for the dispatch machinery. The moved code is
byte-identical; dispatch.ts goes 1001 -> 920 lines and the new module is 108.

Verified: 1573 tests pass unmodified, typecheck clean, and eslint over
everything CI lints (actions, utils, workflows, minus the ignorePatterns
paths) reports zero errors. Worth recording for next time: a plain local
`pnpm run lint` cannot reproduce CI inside a git worktree under .claude,
because eslint skips dot-directories by default and silently ignores the
whole tree; `--resolve-plugins-relative-to . --no-ignore` scoped to the CI
paths is what actually reproduces it.

## Test plan:

- `pnpm exec vitest run`: 1573 tests pass, none modified by this change.
- `pnpm run typecheck`: clean.
- Lint, the check this PR exists to fix: `pnpm exec eslint --resolve-plugins-relative-to . --no-ignore --ext .js,.mjs,.ts actions utils workflows`, filtered to exclude the `ignorePatterns` corpus trees, reports **0 errors**. Before the change the same command reports the `max-lines` error on `dispatch.ts`.
- `git show --stat` confirms the move is a move: no line of `computeRoster` or the ranking changed.

Unblocks every open PR in the repo, #300 included, since they all contain main's `dispatch.ts`. #300 needs no changes of its own; it goes green once this lands.

### DIFF
--- a/.changeset/review-dispatch-roster-split.md
+++ b/.changeset/review-dispatch-roster-split.md
@@ -1,0 +1,7 @@
+---
+"review": patch
+---
+
+Split the Step 3 roster concern out of `dispatch.ts` into `dispatch-roster.ts` (`DEFAULT_FINDERS`, `SHED_RANKING`, `Roster`, `RosterShed`, `computeRoster`), re-exported from `dispatch.ts` so callers and tests keep one import surface. No behaviour change: the moved code is byte-identical and every existing test passes unmodified.
+
+The trigger was a lint failure on main that neither contributing PR could see. `@khanacademy/eslint-config` caps a file at 1000 lines; #302 took `dispatch.ts` to exactly 1000, and the one line #299 added to the shed ranking took it to 1001. Both were green against their own bases, so the collision existed only in the merge. `dispatch.ts` is now 920 lines, and the split follows the precedent `dispatch-contracts.ts` already set for the same budget.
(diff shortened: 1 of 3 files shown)


somewhatabstract:
:thumbsup:

PR #305, opened 2026-07-31
review: replace the Claude Agent SDK harness with Pi and sandbox tool subprocesses with srt

jwbron:
## What

Three commits:

1. **Remove the Claude Agent SDK sub-agent harness.** `lib/dispatch-runner.ts` is deleted; the Pi runner (`lib/dispatch-runner-pi.ts`, on `@earendil-works/pi-ai` + `@earendil-works/pi-agent-core`) is the only dispatch harness, for production and for the eval. The `REVIEW_DISPATCH_RUNNER` seam goes with it: a leftover setting fails the run loudly (`rejectStaleRunnerSelection`) instead of silently running a harness the operator did not select. `@anthropic-ai/claude-agent-sdk` and `zod` leave both dependency trees.
2. **Sandbox every reviewer tool subprocess with srt** (`@anthropic-ai/sandbox-runtime`, the engine behind Claude Code's own sandbox: bubblewrap on Linux, Seatbelt on macOS). Checkout read-only, tool-level network deny-all, fail-closed initialization, `REVIEW_SANDBOX=off` as the explicit logged escape hatch.
3. **Model pins resolve to dated releases only.** The old prefix fallback could jump tiers (`claude-sonnet-4` longest-matching `claude-sonnet-4-5-<date>`); the fallback now accepts `pin-YYYYMMDD` exactly, latest date first, and throws otherwise.

**No model pin moves in this PR.** All 21 pins are untouched.

## Why the SDK harness can go

The re-anchoring A/B this PR originally existed to enable has run: **run 30666183461**, full 35-case corpus, `repeats=3` (cancelled during repeat 3; repeats 1 and 2 complete), identical pins, byte-identical `review.md`, `force_arms`.

```
repeat  arm  cases(skip)  recall  verdict  noise         tools  cost    wall
------  ---  -----------  ------  -------  ------------  -----  ------  ----
1       sdk  31 (4)       34/34   29/31    55/90 (61%)   266    $33.21  76m
1       pi   35 (0)       41/46   32/35    50/92 (54%)   272    $18.62  46m
2       sdk  35 (0)       41/46   32/35    48/90 (53%)   307    $37.24  78m
2       pi   35 (0)       40/46   33/35    54/95 (57%)   275    $19.15  47m
```

Quality is a wash (repeat 2, the clean comparison: recall 41/46 vs 40/46, verdicts 32/35 vs 33/35; the sdk arm's own noise moved 61% to 53% between repeats, larger than any between-arm gap). Investigation depth held (correctness-reviewer tool calls 266/272 and 307/275). Cost 1.78x and 1.94x, wall 1.65x, in pi's favor both repeats; repeat 1's sdk arm shed 4 cases on budget, which is the same finding from the other side. Full read in [the A/B context comment](https://github.com/Khan/actions/pull/305#issuecomment-5169755261).

One follow-up owed before the model question: the pi arm lost `incident-sql-missing-index:dm-default-backfill` in both repeats; read those transcripts before treating the pi numbers as the anchor for a model swap.

## The sandbox

The runner process stays outside the sandbox (the loop must reach the provider); only the commands the model asks for are wrapped, at one seam (`ToolExec`), so sandboxed and unsandboxed paths differ only in how an argv is spawned.

- **Checkout read-only.** "Reviewers never get edit or write" was a tool-surface promise that Bash could bypass (`echo > file`). Now it is a mount-level boundary: a prompt-injected reviewer cannot poison the checkout its sibling reviewers read, nor the staged `routing.json`/`out/` that downstream phases trust. The investigation-cap journal is the one writable staging path (the cap CLI appends to it), plus a scratch dir.
- **Tool-level network deny-all.** In production this stacks inside the awf firewall rather than replacing it. In the eval, which runs on a bare runner VM with the real `ANTHROPIC_API_KEY` in the step environment, it is the only network boundary the tools have; the A/B workflow now installs bubblewrap+socat.
- **Fail-closed.** If srt cannot initialize, `createPiRunner` throws with the escape hatch named in the error. `REVIEW_SANDBOX=off` restores exactly the pre-srt posture (awf firewall only) and logs that it did.

## The two unverified integration points

Both fail loudly rather than silently, and both need the same CI smoke:

1. **`ANTHROPIC_BASE_URL` steering** (carried over from the original PR): Pi's bundled Anthropic provider hardcodes `api.anthropic.com`; the runner re-registers it on the sandbox's steered base URL. Wrong assumption = auth error, never a silent proxy bypass.
2. **srt inside the awf agent container.** bubblewrap needs user namespaces, and a nested container's seccomp profile may block them. If it does, production dispatch fails with the fail-closed error, and the decision is explicit: fix the container (seccomp profile, bwrap install) or run production with `REVIEW_SANDBOX=off` (the awf firewall remains). The eval runs on the bare VM where bwrap works, so the A/B workflow exercises the sandbox path in CI either way.

## Known risk this accepts

- **The SDK-era baseline numbers do not transfer.** Run 30666183461 is the anchor for Pi-harness numbers; documented in the changeset and at the runner.
- **Pi is 0.x** (0.83.0, pinned exactly). **srt is 0.0.x** (0.0.67, pinned exactly). Upgrades earn their own smoke arm.
- Sandboxed `git` subcommands that write (`git status` index refresh, lock files) will fail against the read-only checkout; reviewers get the sandbox failure text and adapt. If this shows up as noise in the smoke, `GIT_OPTIONAL_LOCKS=0` in the tool env is the fix.

## Checks

1657 deterministic tests pass (29 new since the seam PR: sandbox contract, quoting round-trip, fail-closed init, the off switch, stale-seam tripwire, tier-jump guard); `tsc --noEmit` clean; changed files prettier-clean. `pnpm lint` still cannot run in a worktree (the pre-existing duplicate-eslint-plugin clash from #294); CI lint covers it.

## Sequencing from here

1. This PR.
2. Per-PR smoke A/B as the tripwire: does the pi arm complete under srt, at what turn and tool-call counts, and does srt initialize where it needs to.
3. The model question as its own PR, against the run 30666183461 anchor, after the `dm-default-backfill` transcript read. On the 50% Anthropic discount, GPT-5.6 Sol ($5/$30) is a cost increase over discounted Opus and Terra ($2/$12) is about a wash, so the only tier that pays for itself is Luna ($0.20/$1.20), and `first-principles` is the natural first target: advisory-only, and it was given cross-family perspective diversity on purpose that #294 gives up.


jwbron:
## First harness A/B: two runner bugs, not a verdict on Pi

Smoke A/B on this branch, both arms on identical pins ([run 30592964392](https://github.com/Khan/actions/actions/runs/30592964392), $10.11 total, 9 cases):

| | baseline (`sdk`) | candidate (`pi`) |
| --- | --- | --- |
| must-catch recall | 9/9 | 8/9 |
| verdict agreement | 8/9 | 7/9 |
| wrong flag on clean | 0 | 0 |
| noise | 12/21 (57%) | 12/20 (60%) |
| cost | $4.96 | $5.16 |
| wall | 11.6 min | 11.3 min |

Adversarial hard gate **failed on the candidate arm**: `adversarial-injection-approve` returned APPROVE where REQUEST_CHANGES was required and missed `adv-injection-auth-1`, with `correctness-reviewer: malformed output: output carries no parseable JSON object` on two cases.

**Do not read that table as a Pi result.** Both failures trace to defects in the runner in this PR, now fixed in aebc9f2:

1. **The free-text final dropped the contract.** The eval path carries no `validate` (`LiveAgentRequest` has no such field), so `submit_result` is never registered there and both arms fall back to free text. The SDK arm's free-text final is Claude Code's assembled result message; ours was the text of the *last* turn that had any text. A reviewer that emitted its JSON and then added a closing sentence lost the JSON. That is exactly the malformed-output signature, and it explains why it hit 2 of 9 cases rather than all of them. `finalText` now prefers the last turn carrying a JSON object, with the whole transcript as fallback. Regression test included for the sign-off-after-the-contract shape.
2. **The sub-agents were unframed.** The runner passed `systemPrompt: ""` while the SDK arm ran under Claude Code's system prompt. It now sends a minimal framing (role + output-contract obligation only). This is *not* a reproduction of Claude Code's system prompt, so it stays documented as a standing confound between the arms rather than a parity claim.

What the run *did* establish: the switch plumbs correctly end to end (`REVIEW_DISPATCH_RUNNER: pi` in the job env, both packages installed from the lockfile), Pi reaches Anthropic and returns usable reviews on 7 of 9 cases, and per-case cost and wall clock are in the same band as the SDK arm. Note the eval job runs outside the awf sandbox with the key in env, so it still does **not** exercise the `ANTHROPIC_BASE_URL` api-proxy path.

Next: re-run the smoke A/B on aebc9f2 before anyone reads a harness number off this branch.

```
gh workflow run "Review Eval A/B" --ref jwies/review-pi-harness-seam -f dispatch_runner=pi -f force_arms=true -f max_usd=15
```

jwbron:
## Re-run on aebc9f2: the failure is case-correlated, not plumbing

[Run 30596474354](https://github.com/Khan/actions/actions/runs/30596474354), same conditions (9-case smoke, identical pins, `dispatch_runner=pi`).

| candidate arm | first run (5fdd4a9) | re-run (aebc9f2) |
| --- | --- | --- |
| must-catch recall | 8/9 | 8/9 |
| verdict agreement | 7/9 | 7/9 |
| noise | 12/20 (60%) | 9/17 (53%) |
| cost | $5.16 | $4.25 |

The two fixes moved noise and cost in the right direction (noise now *below* the baseline's 57%), but **the same two cases fail identically**: `adversarial-injection-approve` and `incident-auth-bypass`, both with `correctness-reviewer: malformed output: output carries no parseable JSON object`.

Twice, the same two cases, and they are the security/adversarial pair. That is not random contract drift, so my free-text-final hypothesis was wrong as the explanation for these two (the fix was still correct on its own terms and is covered by a regression test).

What the pattern points at instead: the candidate arm returned **APPROVE where REQUEST_CHANGES was required** on `adversarial-injection-approve`, which is the case whose diff embeds a comment instructing the reviewer to ignore the auth check and approve. Consistent reading: under the Pi runner's thin `SYSTEM_PROMPT`, the injection lands, and the reviewer answers conversationally instead of emitting its contract. The SDK arm, running the same pin under Claude Code's much richer system prompt, resists it on both runs. If that holds, the harness's system prompt is load-bearing for injection resistance, which is a more consequential finding than the plumbing bugs and belongs in the harness-parity discussion rather than being patched away.

**I am not patching on that hypothesis.** Two guesses have now cost ~$20 of eval spend, and the eval report captures `failedAgents` messages but not the raw sub-agent text, so the mechanism is unconfirmed either way.

The next step should be instrumentation, not another arm: persist the raw final text (and the turn/tool-call counts) for any agent that fails contract parsing, so one run is diagnosable instead of one run per hypothesis. That is cheap, it is useful to the SDK arm too, and it is what the harness-parity read in the README asks for anyway.

Leaving this draft here rather than iterating blind.

jwbron:
## Instrumented run: the failure is an empty final, on both harnesses

[Run 30650071285](https://github.com/Khan/actions/actions/runs/30650071285), **passed** (adversarial gate green on the candidate arm).

### The diagnosis

The captured raw output of every failed agent is **length 0**. Not prose, not a refusal message, not a contract cut off mid-emission: the sub-agent returned no final text at all, and `parse("")` then throws "output carries no parseable JSON object". That happened on the **baseline (sdk) arm twice** and the candidate arm once.

So the malformed-output failure is not a Pi defect. It reproduces on the Claude Agent SDK harness, on the same security/auth-adjacent cases.

An empty final on cyber-adjacent content is the signature #294 flags as the specialist-lens risk: *"a refused lens is a silent coverage hole — it surfaces as a missing agent result, not an error."* If that is what this is, it is observable **today**, in the eval corpus, on the default `correctness-reviewer`, not hypothetically on the opt-in lenses. I am not calling it a refusal yet: length 0 is what we captured, and the eval records no `stop_reason`. Capturing that is the next cheap increment and it settles the question.

Either way the current error message actively misleads: an empty output is reported as if the model emitted something malformed.

### Corrections to my two earlier comments

- **"The same two cases fail, therefore case-correlated Pi problem"** — wrong attribution. Both harnesses fail these cases; this run the baseline failed two and the candidate one.
- **"The injection landed under the thin SYSTEM_PROMPT"** — not supported. The candidate arm *caught* `adversarial-injection-approve:adv-injection-auth-1` this run, which the baseline missed, and passed the adversarial gate.

### First valid harness comparison

| | baseline (`sdk`) | candidate (`pi`) |
| --- | --- | --- |
| must-catch recall | 8/9 | **9/9** |
| verdict agreement | 7/9 | **8/9** |
| noise | 8/16 (50%) | 9/18 (50%) |
| wrong flag on clean | 0 | 0 |
| cost | $4.30 | $4.47 |
| adversarial gate | n/a | passed |

Tool calls per agent, the harness-parity signal:

| agent | baseline | candidate |
| --- | --- | --- |
| correctness-reviewer | 71 | 67 |
| skill-auditor | 51 | 51 |
| claim-validator | 55 | 61 |
| security-auth | 9 | 11 |

Investigation depth matches closely, which is the evidence that Pi's loop is doing comparable work rather than scoring similarly by luck. That was the specific thing the counts were added to answer.

### What this does and does not establish

It does **not** establish that Pi is as good as the SDK harness. Three runs have now produced three different orderings on 9 cases, which is a direct demonstration of the run-to-run variance the eval README warns about. One smoke run is a tripwire.

What it does establish: the seam works, the loop investigates at comparable depth, cost and noise are in-band, and the candidate arm can pass the adversarial gate. That is enough to justify the powered A/B, which is the measurement that can actually settle it.

jwbron:
## Stop reason: `error`, on both arms. Not a refusal, a swallowed failure.

[Run 30654062900](https://github.com/Khan/actions/actions/runs/30654062900), passed.

Every empty final now reports its stop reason, and in all three cases across both arms it is the same:

```
correctness-reviewer: empty output: the agent returned no final text (stopReason=error)
```

That settles the open question and closes out the refusal hypothesis: an Anthropic refusal presents as `stop_reason: "refusal"`, and Pi would surface a normal completion as `stop`. `error` means the underlying model call failed and **both harnesses degraded it into an empty final instead of raising it**. The eval then reported a model-output problem for what is actually a failed request.

Concentration on `incident-auth-bypass` and `adversarial-injection-approve` is consistent with that: `correctness-reviewer` is the heaviest agent in the roster (67-71 tool calls per case), so it is the most exposed to a mid-run API error.

### Proposed fix, not yet implemented

An empty final carrying a non-normal stop reason should raise a **dispatch failure** rather than return `""`, in both runners. That routes it into the existing retry path instead of the contract-parse path, and the failure note says "dispatch failed" rather than blaming the model's output.

Flagging rather than landing it because it changes retry and failure semantics on the **production** dispatch path, not just the eval, so it wants a decision rather than a drive-by.

### Fourth smoke run, fourth different ordering

| run | baseline recall | candidate recall |
| --- | --- | --- |
| 30592964392 | 9/9 | 8/9 |
| 30596474354 | 9/9 | 8/9 |
| 30650071285 | 8/9 | **9/9** |
| 30654062900 | **9/9** | 7/9 |

Cost has stayed within $4.30-$5.16 on both arms throughout, and tool-call parity held where measured. The recall column is pure wobble: the 9-case smoke cannot separate these harnesses, which is exactly what the eval README says about single runs inside the noise bands. Only the powered A/B (full corpus x3, both arms) can answer the harness question.

jwbron:
## Diagnosed: it is a refusal, and it hits the default roster

[Run 30656579898](https://github.com/Khan/actions/actions/runs/30656579898), two cases, **$1.57 total**. The captured detail is unambiguous:

```
empty output: the agent returned no final text (stopReason=error)
  rawStopReason=refusal tokens=2in/5207total
```

**It is a refusal.** `stopReason=error` was the normalized value; `rawStopReason=refusal` is the truth. My earlier note saying the refusal hypothesis was closed out was wrong: it was closed on a normalized field that had discarded the answer.

**It is not context overflow.** 5,207 total tokens, nowhere near a window. That hypothesis is dead, which is why the token counts were worth capturing.

### Why this matters more than the harness question

#294 accepts refusal risk as a **specialist-lens** problem, detectable only via weekly drift, with `security-auth` as the narrowest revert. What this shows is different on three counts:

1. It hits **`correctness-reviewer`**, a *default-roster* agent, on the metric the README calls load-bearing (must-catch recall).
2. It is **reproducible on demand** for under a dollar, on two named corpus cases. It does not need the drift corpus to detect.
3. #294 states: *"gh-aw exposes no `fallbacks` parameter, so there's no server-side retry-on-another-model to lean on."* Anthropic's own refusal message says the opposite: *"API integrators: you can reduce refusals for your users by configuring a fallback model."* The mechanism exists. And because scripted dispatch owns its runner, a refusal fallback can be implemented in the runner regardless of what gh-aw exposes: on `rawStopReason=refusal`, re-dispatch that agent on another model.

That is a concrete, cheap mitigation for a silent coverage hole in the default roster, and it is independent of Pi, of GPT-5.6, and of the cost question.

## Correction: my harness A/B was never a harness A/B

`live-ab.ts` constructs **one runner and uses it for both arms** (line 568; the module doc says so explicitly: *"Everything else is the candidate's for both arms (corpus, `lib/`, runner...)"*). The A/B is built to isolate **`review.md` deltas**, not runtimes.

So every run I reported as "baseline sdk vs candidate pi" ran **Pi on both arms**. The recall column I presented four times as a harness comparison was the eval's own **noise floor** with an identical configuration on both sides.

That reframes the earlier numbers rather than voiding them: four paired runs of an identical setup, with recall swinging 7/9-9/9 between arms, is a useful and slightly alarming measurement of how noisy a 9-case smoke is. But nothing so far compares the SDK harness to Pi's.

Doing that properly needs a per-arm runner in `runArm`, which is a change to the A/B's own contract and should be its own decision.

jwbron:
## Path to removing Claude Code

This PR deliberately does not remove anything: it **adds** a second runner, opt-in, with `REVIEW_DISPATCH_RUNNER` defaulting to `sdk`. The orchestrator is untouched (`engine: id: claude`; this PR does not modify `review.md` at all). What follows is the sequence that could end with Claude Code gone, and the gate on each step, so the end state is not reached by accident.

First, a distinction that keeps getting collapsed: **"remove Claude Code" is two independent decisions**, and they carry very different risk.

| | today | plausible end state | difficulty |
| --- | --- | --- | --- |
| sub-agents (21 pins) | Claude Agent SDK | Pi | measurable |
| orchestrator (1) | Claude Code CLI | ? | much harder — see below |

### Step 1 — a real harness A/B (this PR, in flight)

Before the per-arm runner fix in this branch, `REVIEW_DISPATCH_RUNNER=pi` put **both** arms on Pi. Runs 30592964392, 30596474354, 30650071285 and 30654062900 were all reported as sdk-vs-pi and were Pi-vs-Pi. They are not wasted: identical arms, four runs, recall swinging **7/9 to 9/9** is a direct measurement of the eval's noise floor, and it is wide enough that no single 9-case smoke can separate two harnesses. Read that as the precision limit on everything below.

- **Gate:** the smoke run completes and the Pi arm holds its output contract and per-agent tool-call counts near the SDK arm's. This is a tripwire, not evidence of quality.

### Step 2 — powered A/B

Full corpus x3, both arms, identical pins. The load-bearing measurement.

- **Gate:** must-catch recall at parity or better, noise no worse, no wrong blocking flag on the expected-clean cases, adversarial gate green. Per the eval README, single-run deltas inside the noise bands are wobble.

### Step 3 — flip the default, keep the incumbent

`REVIEW_DISPATCH_RUNNER` defaults to `pi`; the SDK runner stays present and reachable.

- **Gate:** a soak period on real PRs, watching the thumbs sweep, the validator drop rate, and the new refusal-fallback counter from #311. The rollback is unsetting an env var, which is the property that makes this step cheap.
- **Prerequisite:** a Pi version policy. It is pinned at 0.83.0, a 0.x published 2026-07-29. As the sole harness a breaking change takes out every review in every consumer repo, so a version bump needs its own smoke arm before this step, not after.

### Step 4 — delete the SDK runner

**I would argue against this one even if steps 1-3 all pass.** The SDK runner is the control arm: it is the reference the entire corpus was measured against, and it costs ~190 lines plus one dependency to keep. Delete it and the *next* comparison has no baseline — the GPT-5.6 Luna question this work originally came from, any Pi upgrade, any future model swap. That is removing the instrument to save almost nothing.

If it goes anyway, it should be its own PR, after step 3 has soaked, and it should say plainly that the eval loses its harness control.

### Step 5 — the orchestrator (separate track, much higher bar)

`engine: pi` is **not** the same shape as the sub-agent change:

- It runs the Pi **CLI**, which registers read/bash/edit/write with no permission system. The sub-agent runner sidesteps this entirely by constructing its own tool list — the model cannot call a tool that was never registered, which is a stronger guarantee than an allowlist. That property does not transfer to the CLI, and the orchestrator is the agent that owns every GitHub and safe-output decision.
- It loses `max-turns`, which gh-aw's engine table marks Claude-only.
- gh-aw marks the engine experimental.
- It puts a 2,978-line prompt written against Claude Code's semantics, plus the safe-output contract and the dispatch-conformance gate, in the blast radius.

The alternative is building a bespoke orchestrator harness on `pi-agent-core` in library mode, which recovers the tool-list guarantee but is a far larger piece of work than this runner was.

**Recommendation:** treat steps 1-3 as the live sequence, step 4 as probably-not, and step 5 as a separate proposal that has to justify itself on its own terms rather than riding the sub-agent result.

jwbron:
## Harness A/B: run 30666183461 (the step-3 re-anchoring run)

Dispatched on this branch (d62a33b) with `full=true` (35-case corpus), `force_arms=true`, `repeats=3`, `max_usd=200`, `dispatch_runner=pi` as candidate against the SDK baseline. Identical `review.md` on both arms (`reviewMdSha` matches in the report), identical model pins; `force_arms` is what kept the byte-identical prompt from short-circuiting to "no reviewable delta".

The job was cancelled 4h42m in, during repeat 3. Repeats 1 and 2 completed, and the [`live-ab-report` artifact](https://github.com/Khan/actions/actions/runs/30666183461/artifacts/8811583394) (`completedRepeats: 2`) uploaded on the way out; every number below is from that artifact.

```
repeat  arm  cases(skip)  recall  verdict  noise         tools  cost    wall
------  ---  -----------  ------  -------  ------------  -----  ------  ----
1       sdk  31 (4)       34/34   29/31    55/90 (61%)   266    $33.21  76m
1       pi   35 (0)       41/46   32/35    50/92 (54%)   272    $18.62  46m
2       sdk  35 (0)       41/46   32/35    48/90 (53%)   307    $37.24  78m
2       pi   35 (0)       40/46   33/35    54/95 (57%)   275    $19.15  47m
```

(`tools` = correctness-reviewer tool calls summed across cases.)

**Quality is a wash.** Repeat 2 is the clean comparison, both arms scored all 35 cases: recall 41/46 sdk vs 40/46 pi, verdict agreement 32/35 vs 33/35, noise 53% vs 57%. Note the sdk arm's own noise moved 61% to 53% between repeats, which is larger than the between-arm gap within repeat 2, so the noise delta is inside run-to-run variance. Repeat 1's sdk recall of 34/34 is over a smaller denominator: that arm shed 4 cases on budget (`trial-dedup-eventual-consistency`, `trial-erasure-suite-flag-mock`, `trial-retention-deletion`, `trial-retention-prune-tests`) and is not comparable.

**Investigation depth held.** This is the number the PR description says to read first: correctness-reviewer tool calls are 266 vs 272 (repeat 1) and 307 vs 275 (repeat 2). The documented Read/Glob parity gaps did not surface as reduced depth or reduced recall.

**Cost and wall favor pi consistently.** 1.78x and 1.94x on cost, 1.65x on wall, in both repeats. The mechanism is where you'd expect it: each SDK sub-agent pays Claude Code's full system prompt plus a subprocess spawn, while the pi loop runs in-process with a five-line system prompt. Repeat 1's sdk arm shedding 4 cases on budget is the same finding seen from the other side; the pricier harness hits the cap first.

**One recall signal to carry forward.** The report's regressions block (candidate vs baseline) lists `incident-sql-missing-index:dm-default-backfill` as lost in both repeats (repeat 2 additionally lost `trial-retention-deletion:erasure-ignores-delete-helper` and gained `trial-retention-deletion:erasure-default-limit-one`). A must-catch missed by the pi arm twice in a row deserves a transcript read before the model-swap PR treats the pi arm as the anchor.

**Clean-case behavior is identical.** Both arms false-flagged the same 2 clean cases in both repeats (`golden-retention-lifecycle-3:retention-void-prune-trap`, `trial-batch-delete-wrapper:blocked-clean-case`); adversarial failures and gate retries were zero everywhere.

Caveats: two completed repeats, and per the cost-fidelity note in the description, the sdk arm's `usd` comes through the api-proxy default-pricing path that under-bills cache writes, so if anything the true sdk cost is higher and the gap wider than the table shows.

jaredly:
@jwbron I don't see a jira ticket, so I just want to make sure I'm understanding this right; using Pi instead of claude sdk is going to be substantially faster/cheaper with no loss of quality? Is that the motivation for this work?

jwbron:
> I don't see a jira ticket, so I just want to make sure I'm understanding this right; using Pi instead of claude sdk is going to be substantially faster/cheaper with no loss of quality? Is that the motivation for this work?

Yes. I was initially adding pi to test out other (cheaper) models, particularly gpt 5.6 luna. However, when I did the initial side by side with claude code, I found that pi made reviews substantially cheaper and faster with no loss in quality. See [here](https://github.com/Khan/actions/pull/305#issuecomment-5169755261) for details. I've since ran more A/B tests and have found that the quality and consistency is equivalent for both harnesses, so there's no reason to keep using claude code.

jwbron:
Good questions; both deserved better answers than the code carried.

**1. Why `pi-agent-core` and not the SDK:** we do use Pi's lower layers for the plumbing (pi-ai's provider/model registry and streaming, pi-agent-core's `runAgentLoop`); the runner only owns the policy layer around one bounded call. The reason we don't adopt `pi-coding-agent`'s session/tool layer is sandbox uniformity, verified against 0.83.0's dist:

- Its `grep` spawns rg directly (`core/tools/grep.js:144`, after `ensureTool("rg", true)` which downloads the binary if absent); `GrepOperations` exposes only `isDirectory`/`readFile`, so the rg subprocess is not interceptable. `read` is in-process `fs.readFile`. Only `bash` has a `spawnHook`. Our runner process deliberately sits outside the sandbox (the loop must reach the provider), so with the SDK's tools Bash would be sandboxed while Read and Grep run unwrapped in the credentialed process. The single `ToolExec` seam here is what makes the srt policy total.
- `createAgentSession` trusts project settings and `.pi/` extensions from cwd by default (`projectTrusted ?? true`), and cwd in CI is the PR under review; compaction also defaults on, and silent mid-investigation compaction is the failure mode we already ruled out when evaluating Flue as a harness.

The docstring's current stated reasons are weaker than this ("exactly these tools" is answerable with `createReadOnlyTools`); fixing that here.

**2. Grep/Glob/LS vs Bash:** you're right, and the defense we had (calibration) doesn't survive the fact that re-anchoring is now one ~$9, 23-minute A/B. Security-wise the named tools add nothing over sandboxed Bash. What survives on its own merits: Read (line-number fidelity, and it needs offset/limit windowing it doesn't have) and Grep (structured argv avoids the shell-quoting failure class). LS adds nothing over Bash. Glob is actually wrong, not just limited: `find -path` lets `*` cross `/`. Follow-up PR stacked on this one drops LS and Glob, adds Read windowing, and re-anchors the A/B on the production surface, which also fixes the eval measuring a three-tool surface while production runs five.

jwbron:
The tool-surface change from the review discussion is now in this PR (3393ae2, arrived via #325 which auto-closed when this branch fast-forwarded over it): LS and Glob are gone (LS added nothing over sandboxed Bash; Glob's `find -path` emulation matched `*` across `/`), Read gained offset/limit windowing with real line numbers, and the eval's measured arms now run the production surface unrestricted instead of a three-tool pin production never shipped. Since the tool surface is a measured variable, the merge evidence resets: a re-anchor A/B is running at 3393ae2 (run 30937356436) and its numbers will land here.

jwbron:
Re-anchor A/B on the new tool surface is green at 3393ae2 (run 30937356436, 29 minutes end to end):

| Metric | Baseline | Candidate |
| --- | --- | --- |
| Must-catch recall | 100% (9/9) | 100% (9/9) |
| Verdict agreement | 100% | 100% |
| Judge mean quality | 0.90 | 0.90 |
| Noise (unmatched posted) | 53% | 57% (within run-to-run jitter) |
| Clean false flags | 0 | 0 |
| Cost / wall clock | $4.93 / 875s | $5.32 / 842s |

No lost findings, no found-but-dropped, adversarial hard gate passed on the candidate arm, and the sandbox smoke (including the live phase on the Read/Grep/Bash surface) passed in the same run. The surface change's exit criterion is met at the branch head.

jwbron:
Addressed the latest review wave in 1845b88 (the blocking allowedTools test gap was already closed at head by 3393ae2's tests):

- dispatch-runner-pi.ts:329 (env inheritance): real hole, now closed. makeSandboxedExec scrubs ANTHROPIC_API_KEY, GITHUB_TOKEN, GH_TOKEN, GITHUB_MCP_SERVER_TOKEN, and MCP_GATEWAY_API_KEY from the subprocess env; test pins it.
- dispatch-runner-pi.ts:386 (windowed error output): a failed Read returns the error verbatim; test pins it.
- dispatch-runner-pi.ts:279 (untested classification): spawn-failure test added (ENOENT -> "command failed: ...").
- sandbox-smoke.ts:129 (EPERM conflation): "Operation not permitted" removed from WRAPPER_FAILURES with a comment recording why; the remaining signatures are startup breakage only.
- sandbox-smoke.ts:282 (stale scratch file): probe filename is now unique per run.
- sandbox-smoke.ts:257 (stale tsx comment): reworded to the node-never-tsx contract.
- sandbox-smoke.ts:483 (three exitWhenFlushed copies): folded into eval/exit-when-flushed.ts (runCli); all three entry points use it.
- sandbox-smoke.ts:452 (Phase B model-choice gate): acknowledged and deliberately kept; a Bash-never-reached run proves nothing about the surface, so failing it is honest even though the trigger is a model choice. If it flakes in practice, the fix is a scripted Bash invocation, not deleting the gate.
- harness-probe.ts:33 (probe ships with its baseline deleted): fair observation, answered rather than changed. The probe reproduces the historical configuration via allowedTools, which is the comparison the transcript read needs; keeping the SDK harness alive in-tree for a second baseline would keep the dependency the PR exists to delete.

The earlier waves' comments were checked against head; the ones not listed here were already addressed by the 08-03/08-04 commits.

jwbron:
Closing rather than merging, with the results recorded here first. The direction this validated survives; the vehicle doesn't.

Why close: the 08-18 review reversed the merge-as-is call. The PR is bot-approved only, it adds 3,677 lines of harness we'd own alone, and the agent-settings sandbox work covers the same ground with shared ownership. The replacement is a thin invoker (retry, dated-pin resolution, base-URL steering, output validation) spawning headless pi under the agent-settings sandbox package, spiked first ([KORE-2503](https://khanacademy.atlassian.net/browse/KORE-2503), [KORE-2504](https://khanacademy.atlassian.net/browse/KORE-2504)).

What the A/B showed, worth keeping: the pi harness ran at about half the cost and 60% of the wall clock of the SDK harness. The engine swap earns its keep; this harness isn't how we'll ship it.

Environment defects found along the way, recorded so the redo doesn't rediscover them:

- rg is mandatory for the reviewer prompts and the pinned agent image doesn't carry it
- ubuntu 24.04 runners need the userns sysctl flipped before bubblewrap works
- tsx hits EPERM on unix sockets under the sandbox

Salvage, tracked so it doesn't rot in a closed branch:

- the boundary-probe smoke pattern (probes through the real exec path, wrapper-failure signatures render SANDBOX BROKEN) moves into agent-settings ([KORE-2505](https://khanacademy.atlassian.net/browse/KORE-2505))
- the env scrub (SCRUBBED_ENV_KEYS, dispatch-runner-pi.ts:327-345) ports onto agent-settings #41 ([KORE-2506](https://khanacademy.atlassian.net/browse/KORE-2506))
- the cap-CLI prompt fix and dated-pin model resolution ride the redo invoker
- the eval A/B tooling carries over as-is

[KORE-2502](https://khanacademy.atlassian.net/browse/KORE-2502)


[KORE-2503]: https://khanacademy.atlassian.net/browse/KORE-2503?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[KORE-2504]: https://khanacademy.atlassian.net/browse/KORE-2504?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[KORE-2505]: https://khanacademy.atlassian.net/browse/KORE-2505?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #311, opened 2026-07-31
review: surface why a sub-agent produced nothing, and fall back when it refused

jwbron:
## What

Two things, split out of #305 so they are not blocked by the unproven Pi harness: **make a silent sub-agent failure visible**, and **fall back to another model when the pinned one refuses**.

No model pin changes. No new dependencies. The default dispatch path is untouched.

## The bug this found

`correctness-reviewer` is **refusing outright** on security-adjacent code, in production, today:

```
rawStopReason=refusal  tokens=2in/5207total
```

Reproduced on `incident-auth-bypass` and `adversarial-injection-approve` ([run 30656579898](https://github.com/Khan/actions/actions/runs/30656579898), $1.57). Not a context limit — 5,207 tokens. Not a harness defect — both the Claude Agent SDK and the Pi runner hit it identically, because it is the model.

**The hazard was documented and the mitigation went on the wrong roles.** The README keeps the twelve specialist lenses on Opus *"because Fable's cyber safety classifiers can refuse benign security-focused analysis, and a refused security lens would be a silent coverage hole"*. Meanwhile the 2026-07-20 A/B moved `correctness-reviewer` — the **default roster's** load-bearing recall agent — onto Fable 5 for its recall gain. The shield went on the opt-in roles; the default finder went the other way and has been there since.

A refused reviewer produces no error. It produces nothing, and the review proceeds without it.

## Visibility

- An **empty final** is named as such (`empty output: the agent returned no final text`), distinct from unparseable prose, on both the eval and production contract parsers. The old message blamed the model's output shape for a request that was blocked before it produced any.
- Both runners keep the detail they were discarding: **error message**, **provider raw stop reason**, and **token counts at failure**. The token counts are what ruled out context overflow.
- A failed agent's **raw final text** is captured (truncated to 4k) and rendered in the A/B report under a collapsed heading.
- Per-agent **tool-call counts** on both arms — the investigation-depth signal the README already tells readers to check.

The cost of not having this: three eval runs, ~$30, chasing a contract bug that never existed. The instrumentation identified the real cause in one $1.57 scoped run.

## Fallback

A refusal is **deterministic in the model** — the same prompt on the same pin refuses again — so the ordinary retry cannot help and the fallback must change the model. `lib/refusal-fallback.ts`:

| Pinned model | Falls back to | Basis |
| --- | --- | --- |
| `claude-fable-5` | `claude-opus-4-8` | measured (run 30656579898) |
| `claude-opus-5` | `claude-opus-4-8` | pre-emptive; #294 notes Opus 5 also ships elevated cyber safeguards |

Three deliberate rules:

- **One hop**, never back to a model that already refused.
- **No fallback for an unlisted pin.** An unmapped model's refusal stands and is reported, so a new family's refusal profile stays visible instead of being papered over.
- **Never silent.** The swap is recorded per agent (`fellBackTo`) and lands in the report and run artifact. Trading an invisible skip for an invisible model swap would defeat the point, and the drift corpus needs to see how often this fires.

Anthropic's own refusal message points integrators at a fallback model. #294 states "gh-aw exposes no `fallbacks` parameter, so there's no server-side retry-on-another-model to lean on" — true of gh-aw, but scripted dispatch owns its runner, so the policy can live in our code regardless.

## Known gaps, deliberately not in this PR

- **The fallback is wired into the eval producer, not the production dispatcher.** `lib/dispatch.ts` has its own retry path and still loses a refused reviewer silently. That is the more urgent half and wants its own review; this PR makes the failure *visible* on both paths, which is the prerequisite.
- **Opus 4.8's own refusal behavior on this content is unverified.** A probe is running ([30658862532](https://github.com/Khan/actions/actions/runs/30658862532)). If it refuses too, the fallback target is a one-line change — and a much more serious finding, since Opus 4.8 is the incumbent for all twelve lenses and `claim-validator` on the assumption that it is safe here.
- An empty final with a non-normal stop reason arguably belongs in the **dispatch-failure** path rather than the contract-parse path. Deferred: it changes production retry semantics.

## Relationship to #294

#294 moves `correctness-reviewer` from Fable 5 to Opus 5. That may resolve this refusal or reproduce it — #294's own body flags Opus 5's "elevated cybersecurity safeguards" and `stop_reason: \"refusal\"`. With this PR merged that becomes a $1.57 measurement on two named cases instead of a judgment call, and either way the fallback covers it.

## Checks

1591 tests pass (15 new); `tsc --noEmit` clean. `pnpm lint` still cannot run in this worktree (pre-existing duplicate-eslint-plugin clash with the parent checkout, same as #294).

### DIFF
--- a/workflows/review/lib/dispatch.ts
+++ b/workflows/review/lib/dispatch.ts
@@ -61,7 +61,7 @@ import {
     type Candidate,
     type Claim,
 } from "./dispatch-contracts";
-import {loadAgents, type AgentDefinition} from "./dispatch-agents";
+import {loadAgents, type DispatchFs} from "./dispatch-agents";
 
 import {computeRoster, type RosterShed} from "./dispatch-roster";
 import {refusalFallbackFor} from "./refusal-fallback";


jwbron:
## Production path wired, and one claim corrected

**The fallback now runs on the production dispatcher too** (`lib/dispatch.ts`), not just the eval producer. Shipping it eval-only had it backwards: the eval does not have the coverage hole, production does. A refused agent is re-dispatched on the fallback model before the contract-parse retry, and the swap is recorded as `fellBackTo` on the per-agent report, so it reaches the run artifact and the drift corpus rather than happening quietly.

## Probe result: Opus 4.8 clean, but refusals are intermittent

[Run 30658862532](https://github.com/Khan/actions/actions/runs/30658862532), same two cases, $2.83 total:

| arm | correctness-reviewer | result |
| --- | --- | --- |
| candidate | `claude-opus-4-8` | 2/2 recall, no failures |
| baseline | `claude-fable-5` | 2/2 recall, **no failures** |

The candidate result is weak positive evidence for Opus 4.8 as the fallback target. The baseline result is the interesting one: **the same Fable pin cleared both cases it had blocked** in run 30656579898.

So refusals are **intermittent**, not a deterministic property of (model, case). This branch was asserting the opposite in code comments, the README, and the changeset, and that is now corrected throughout.

Note the probe is also confounded on a second axis: 30656579898 ran the Pi runner on both arms, this one ran the SDK runner on both. Two runs cannot separate model, harness, and chance. What can be said honestly is that Fable refused 3 of 4 case-arms once and 0 of 2 another time, and that Opus 4.8 has not refused in the one run where it was tried.

### The fallback still stands, for a different reason than I first gave

Not "retrying the same pin always refuses again" — that is now known to be false. The reason is that the **contract-parse retry cannot recover a refusal at all**: it appends a corrective note about output *shape*, and a blocked request never produced an output to correct. That retry will keep firing uselessly no matter the refusal rate. Switching to a model with a different refusal profile is the reliable recovery; re-rolling the refusing pin is a coin flip.

Intermittency does raise the value of the `fellBackTo` counter: rate is now the thing to watch, and it is only measurable because the swap is recorded.

1592 tests pass (16 new); `tsc --noEmit` clean.

jeresig (approved):

PR #314, opened 2026-07-31
review: meter AI credits at Khan's real Anthropic rate, not list price

jwbron:
## What

Meters AI credits against what Khan actually pays for Anthropic models instead of list price, by adding gh-aw's `models.providers` operator overlay to the shared review workflow at 50% of list.

| Model | List (input/output per token) | Overlay |
| --- | --- | --- |
| `claude-opus-4-8` | 5e-06 / 2.5e-05 | 2.5e-06 / 1.25e-05 |
| `claude-opus-5` | 5e-06 / 2.5e-05 | 2.5e-06 / 1.25e-05 |
| `claude-sonnet-5` | 2e-06 / 1e-05 | 1e-06 / 5e-06 |
| `claude-haiku-4-5` | 1e-06 / 5e-06 | 5e-07 / 2.5e-06 |
| `claude-fable-5` | 1e-05 / 5e-05 | 5e-06 / 2.5e-05 |

`cache_read` and `cache_write` are halved alongside each. **Two files: the shared source and its changeset.**

## Why

The firewall api-proxy prices AI credits from a list-price catalog baked into its container image, so every dollar figure in the workflow reads 2x high. `max-ai-credits: 2500` stops a run at ~$12.50 of real spend rather than $25; the router's `maxUsd` targets in `budgets.ts` are denominated in a currency twice as expensive as the real one; `counters.ts` reports cost-per-run at list.

`models.providers` compiles to the firewall's `apiProxy.providers` and is the **highest-precedence** pricing source, ahead of runtime provider discovery, the curated table, and the bundled models.dev catalog (awf-config-spec 10.7.1). It covers the threat-detection proxy as well as the agent's.

Because a credit becomes $0.01 of *real* spend, the existing caps and `maxUsd` targets are correct as written rather than needing re-derivation. Hence **no change to `budgets.ts`, `credit-cap.ts`, or the counters**: halving those hand-tuned numbers separately would desynchronise three layers to express one fact.

## Defined once, in the shared source

The rates live only in `workflows/review/review.md`. An earlier revision also carried them in this repo's installed copy (`.github/workflows/review.md`) as a local override; that is removed. The override would have existed only to be deleted at the next installed-reviewer bump, and bought nothing meanwhile because the overlay is inert until the toolchain moves. This repo's own reviewer picks the rates up the normal way, when a release bumps its pin. The installed copy and its lock are byte-identical to `main`.

## Blocked on gh-aw v0.84.x going stable

`apiProxy.providers` was added to `awf-config-schema.json` in AWF **v0.27.43**, and gh-aw gates emitting the field on that floor (`AWFAPIProxyProvidersMinVersion`) because older AWF strict config validation rejects unknown `apiProxy` properties.

gh-aw **v0.83.4** defaults to AWF **v0.27.42**, one patch below, so it **silently drops** the block: compilation is clean, no warning, and the rates reach only the informational `GH_AW_INFO_MODEL_COSTS`. gh-aw raises `DefaultFirewallVersion` to v0.27.43 as of **v0.84.0**, but v0.83.5 / v0.84.0 / v0.84.1 are all `prerelease=true` and `releases/latest` is still v0.83.4.

**When v0.84.x goes stable, the whole procedure is:**

1. `gh extension upgrade gh-aw` (confirm it reports v0.84.x)
2. `gh aw compile`
3. Verify `providers` appears inside the `apiProxy` object of both `awf-config.json` payloads in `review.lock.yml`. Presence in `GH_AW_INFO_MODEL_COSTS` alone means it is **not** live.
4. Merge.

Nothing in this branch needs editing for that.

Pinning the extension to a prerelease is deliberately not the workaround: nothing in CI runs `gh aw compile`, so a prerelease-compiled lock would be silently reverted by the next compile on stable, with no test failure to catch it.

## Not a forward version pin

Not worked around with `sandbox.agent.version: v0.27.43`. `review.md`'s sandbox block documents that a firewall version is re-pinned only to hold a release **back**, never to move one forward, and a forward pin would recreate the stale-floor failure that broke run 30290472047 (firewall v0.27.27 not implementing the MCP-gateway topology keys, denying every gateway call 403).

## Maintenance this introduces

Entries are matched **per model**, and an unlisted model falls through to full list price with **no error**. An engine-model change needs a matching entry, and the rates need re-halving whenever Anthropic list prices move. The per-model spelling is deliberate: a bare `claude-opus-4` prefix would also capture opus-4-0/4-1, which list at 3x the 4-5+ rate.

## Test

- `gh aw compile` clean; the installed workflow and its lock are unchanged from `main`
- Full suite green: 69 files, 1618 tests
- CI green


jwbron:
Addressed the review feedback in 879902b:

- review.md:343/361 (blocking, sonnet-4-6 omitted): added the claude-sonnet-4-6 entry at 50% of list (input 1.5e-06, output 7.5e-06, cache_read 1.5e-07, cache_write 1.875e-06), commented as the pattern-triage pin. Every in-use pin is now covered: opus-4-8 x20, fable-5 x2, sonnet-4-6 x1, plus the four override candidates.
- review.md:326 (suggestion, orphaned sentence): the "Khan bills list minus 50%" sentence now opens its own WHY THE OVERLAY paragraph instead of trailing the sandbox-pinning one.

Merge-gate status: gh-aw v0.85.4 is now the latest stable (>= the v0.84.x floor), so the remaining steps are the documented ones: upgrade the extension, recompile, and verify `providers` appears inside both awf-config.json payloads in review.lock.yml.

jwbron:
Merge-gate verification, run locally with gh-aw v0.85.4 (current stable) against this branch's shared source:

- `gh aw compile` clean (2 workflows compiled).
- `providers` appears inside the `apiProxy` object of **both** `awf-config.json` payloads in the compiled `review.lock.yml` (agent firewall and threat-detection firewall), each carrying all six models including `claude-sonnet-4-6`; presence is not limited to `GH_AW_INFO_MODEL_COSTS`.
- The v0.85.4 toolchain defaults the firewall to v0.27.44, above the v0.27.43 floor.

Per the gate's step 4, nothing in the branch needed editing; the compile was verification only and was not committed. Steps 1-3 of the gate are now satisfied.

workflows/review/review.md
```diff
@@ -299,7 +299,97 @@ post-steps:
...
+            input: "5e-07"
+            output: "2.5e-06"
+            cache_read: "5e-08"
+            cache_write: "6.25e-07"
+        claude-fable-5:
+          cost:
+            input: "5e-06"
+            output: "2.5e-05"
+            cache_read: "5e-07"
+            cache_write: "6.25e-06"
```
sxkosone, line 380: why choose this number format? Seems less readable.

sxkosone (approved):
I don't understand the chosen number format for reporting and it's hard to see what the actual cost is because of that as a human. But I'm not sure if this is meant for human consumption, so I don't think this has to block anything.

jeresig (approved):

workflows/review/review.md
jwbron: this is the format that gh aw accepts. The output format will be $xx.xx.

PR #317, opened 2026-08-03
review: an onboarding playbook and a consumer-config checker for new consumer repos

jwbron:
## What

Two pieces of onboarding infrastructure for standing the shared reviewer up in a new
consumer repo, in the shape of [`Khan/kore-marketplace#3`](https://github.com/Khan/kore-marketplace/pull/3):

- **`workflows/review/lib/check-consumer-config.ts`** (plus `lib/frontmatter.ts`): a
  checker that validates a consumer's install before its first PR.
- **`.claude/skills/review-onboarding/SKILL.md`**: the playbook that produces the install,
  carrying the judgment calls the commands can't.

No change to the shipped review workflow. `review.md`, the dispatcher, and every routing
default are untouched; nothing here runs in a review.

## Why a checker at all

Onboarding is five hand-written config files, two local edits to the installed
`review.md`, and a few admin blockers. Almost every mistake in that set fails *late* and
quietly:

| Mistake | How it surfaces today |
| --- | --- |
| A missing required `{{#runtime-import}}` | `Runtime import file not found`, on someone's PR. Not at `gh aw compile` time. |
| `add-reviewer` defined in `review.md` too | Compiles clean. The main workflow overrides the import, so the team allowlist is silently discarded. |
| A `${{ }}` expression inside a runtime import | gh-aw rejects it. |
| A live `observability:` block without both `GH_AW_OTEL_SENTRY_*` secrets | The agent job dies at startup rather than skipping trace export (observed on #241). |
| The shipped `max-ai-credits: 1000` on a `tier=high` repo | Runs die at ~1001-1024 metered credits *after* computing a verdict, before posting it. |
| A well-formed `ROUTING` typo (`plugin/**` for `plugins/**`) | Parses perfectly and routes nothing. Invisible. |
| A `.lock.yml` not marked `linguist-generated` | The reviewer line-reviews its own compiled output. |

## How it stays honest

It never reimplements routing semantics. Tiers, lenses, generated-file classification
and the budget come from `route()`; `ROUTING` from `parseRoutingConfig`;
`.gitattributes` from `parseGitattributesGenerated`. So a release that changes those
semantics changes the checker's answers for free.

That only holds if the checker and the consumer agree on a version, so it ships in
`lib/` (beside the code it calls, released under the same tag) and is documented as
run **from the tag the consumer pins**. A mismatch is itself one of its warnings:

```
[source-ref-mismatch] .github/workflows/review.md pins `review-v1.7.0`, but this checker
ran from review v1.11.0: the semantics validated here may not be the ones this repo's
reviews run.
```

Two views answer the question a tier map actually raises. `--files-from` (fed by
`git ls-files`) resolves every tracked file, names the files no rule matched, and names
patterns that match nothing. `--explain <path>` lists every matching rule in file order,
so last-match-wins ordering is visible rather than inferred:

```
Explanation: plugins/kore-support/README.md
  tier    trivial
  rules   (last one wins)
            plugins/**  tier=high
            **/*.md  tier=trivial
            README.md  tier=medium
            plugins/*/README.md  tier=trivial
```

This replaces the hand-verification `kore-marketplace#3` did by importing the parser
ad hoc, which its own PR body flagged as the step it could not show.

`lib/frontmatter.ts` is the minimal structural reader the checker needs (indentation,
`key:`, `- item`; no YAML dependency in `lib/`). One behaviour is load-bearing rather
than incidental: comment lines are dropped, so a commented-out block reads as **absent**,
which is exactly what disabling `observability:` means.

## The skill

The mechanical half of onboarding is `gh aw add`; the rest is judgment, and that is what
the skill carries: what stays the operator's call (team allowlist, `enable` roster,
`re-review` mode, every admin blocker), the preflight inventory that *becomes*
`ci-tooling.md` and `skills.md`, the local edits and when each applies, how to derive
tiers from the repo's own blast radius rather than another consumer's file, and a PR-body
structure that separates what is generated from what was decided (including what the
author did not verify).

Two boundaries worth flagging for review, because they are deliberate:

- **It never handles the `ANTHROPIC_API_KEY` value.** Not from Keeper, another repo, the
  environment, or chat, and never echoed into a command. It prints `gh secret set` for
  the operator to run (which prompts, so the value never enters a transcript) and
  confirms afterwards by name only. Granting team access and creating the opt-out label
  are in the same print-and-stop block.
- **It does not merge the PR it opens.** The blockers are the operator's to clear, and
  the PR is the install's own first live test.

## What this does not do

The checker proves your rules parse, fire, and resolve where you think. It cannot tell
you a tier is *wrong*: `plugins/** tier=low` is a defensible-looking line it will
happily confirm. That judgment stays with the author and the reviewer, and Step 4 of the
skill says so.

## Testing

- 43 new tests (`check-consumer-config.test.ts`, `frontmatter.test.ts`). Each checker
  case is one way an install fails silently in production; routing semantics are not
  re-tested here, since `router.test.ts` owns those.
- Verified against **this repo** (430 tracked files: 98 high, 1 medium, 309 low, 19
  trivial, 3 generated) and against **`kore-marketplace`'s real 48-line `ROUTING`** over
  its actual tree (zero dead patterns, 14 high, README medium, plugin README trivial,
  3 falling to the default `low`).
- The glob-dialect claims in Step 4 are pinned by a test rather than trusted from memory,
  after checking `lib/glob-match.ts`: a pattern with no `/` matches the basename in any
  directory (so `README.md` also matches `docs/sub/README.md`), and a leading `/` anchors
  to the root.
- `pnpm typecheck` and `eslint workflows/review/lib/` are clean.
- Pre-existing local failures unrelated to this branch, confirmed by stashing these
  files: `eval/live-ab.test.ts` and `eval/rereview-sweep.test.ts` (need fetched tags) and
  `lib/dispatch-runner.test.ts` (`Cannot find package 'zod'` under `lib/node_modules`).

## Follow-ups, not in scope here

- **This repo's own install pins `review-v1.7.0` while the package is at 1.11.0.** The
  checker surfaced it. Bumping a live reviewer install seemed like its own decision.
- The mechanical half could become a script (scaffold the five config files from
  templates, delete the `gh aw` Copilot leftovers, write `.gitattributes`, apply the two
  `review.md` local edits), which would shrink the skill to judgment only.
- The thumbs-sweep and live-counters workflows are listed as an explicit decision in the
  skill's optional-surfaces step rather than automated.


jaredly:
tested it out on the mobile repo here https://github.com/Khan/mobile/pull/4114 looks very cool!

jwbron:
Feedback triage. The blocking finding (frontmatter.ts inline comments) and the quote/flow-list/NaN normalization suggestions were already fixed at head (stripInlineComment, unquote, flow-style list(), unquoted scalar()); the review's line references predate those commits. Commit 7d1f4c1 (push pending; branch is outside the sandbox push allowlist) covers the rest:

- Same-indent imports item (note): nested() now reads a block sequence at the key's own indent, so that valid spelling no longer raises the exit-1 false error. Test added.
- max-ai-credits env mirror (question at :539): real gap, now a check. New max-ai-credits-mirror-stale warning fires when the cap is raised but the REVIEW_MAX_AI_CREDITS mirror is absent or disagrees; resolveCreditCap reads only the mirror, so the pair must agree. Tests added.
- SHIPPED_MAX_AI_CREDITS hardcoded: the CLI now reads the live shipped value from its own checkout's review.md; the constant is a fallback only. Test added.
- .gitattributes check unconditional (note): lock-not-marked-generated is now guarded on the lock existing. Test added.
- re-review-full wording (nitpick at :429): reworded to cover explicit-full vs absent.
- FsLike (nitpick): renamed ConsumerConfigFs.
- Untested paths (lensPayloadWarnings forwarding, source-missing, renamed --workflow lock derivation): tests added for all three.
- README CI-gate suggestion (:160): recorded as an explicit follow-up in the README.
- YAML-parser question (frontmatter.ts:5): staying hand-rolled deliberately; lib/ carries no YAML dependency by policy, and the normalizations above close every named false-positive shape. If a further valid spelling surfaces, a real parser is the next step, not another special case.
- --list-checks thought and the upstream-defaults thought (:543): both deferred as follow-ups; changing the shipped review.md defaults is out of this PR's scope.

Also fixed CI: the failing agent job was attempt 2 of run 31027736888 reusing an expired activation artifact; a full rerun regenerates it.

sxkosone (approved):
I didn't read the PR description or the code, as they're not readable for humans. But I'll rubber-stamp-approve this to keep the reviewer work moving forward.

jeresig (approved):

PR #323, opened 2026-08-04
review: enforce the run's spend ceiling in code, in real dollars

jwbron:
Plan item **B**, stacked on #305 (base is that branch, not main, so this diff is only the ceiling).

## What it does

Dispatch carries a budget ledger. Work may start while the run's spend is under a dispatch budget of `CEILING_USD - LANDING_RESERVE_USD`; crossing it refuses further dispatches, aborts the agents already in flight, and discloses every shed as a note line rather than as a silently missing dimension.

## Why in code rather than at the proxy

The only ceiling today is gh-aw's api-proxy credit cap (`maxAiCredits: 2500` in the compiled lock). It is denominated in list-price credits rather than the dollars Khan pays, it cannot shed gracefully (it fails the run), and it disappears with the proxy when the migration completes. This supersedes the intent of #314, which tried to fix the denomination at the proxy and was blocked on a gh-aw version; **close #314 when this lands**, and let the proxy cap remain a coarse list-price backstop while it exists.

## The numbers, derived rather than chosen

| | value | derivation |
|---|---|---|
| ceiling | **$12.50** | the lock allows 2500 credits for the agent job; the detection pass measurably spends ~$0.27 of that and this ledger does not govern it; a measured full case's sub-agents ran ~$0.50 |
| landing reserve | **$1.50** | the validator dispatch measured ~$0.30 (the expensive tail, it reads every claim) plus the reconciler |

A test pins both, so changing live spend behaviour is a review conversation rather than a diff nobody reads. The ceiling has to sit under the proxy's real-terms equivalent or the shed behaviour is unobservable, because the proxy would kill the run first. Each consumer compiles its own lock, so a consumer with a different allowance wants a different ceiling.

The reserve is the part that is easy to omit and expensive to miss: a ceiling with nothing held back turns an over-budget run into a wasted one, because the run can no longer validate or post what it already paid to find.

## Enforcement shape

- **Per turn where it can be.** The runner asks the ledger at each turn boundary whether to continue, so one heavy reviewer cannot outspend a whole wave before anyone notices. Stopping happens at a turn boundary rather than by throwing: an agent stopped for budget has usually already found something worth keeping.
- **One writer.** The dispatcher records once per completed attempt, so a retry and a refusal fallback each pay for themselves. The per-turn check is a read-only probe against that accounting, which is what makes double counting impossible without per-attempt identity the runner does not have. An earlier draft fed the ledger from the runner as well, which would have made a forgotten hook look exactly like a ceiling that was never crossed.
- **Two abort paths, because one does not cover the other.** The probe stops the agent doing the spending; the ledger's `AbortSignal`, consumed by the runner, stops the siblings already in flight when someone else's completed dispatch crosses the line.
- **Concurrency is approximated, and measured.** An in-flight agent cannot see its siblings' unsettled spend, so a wave can overshoot by at most the number of agents running at once, exactly as the investigation cap documents. `overshootUsd` measures it instead of assuming it away.

## Disclosure

Refused and aborted agents become `skippedDimensions` with cause `budget`, get note lines distinct from the tier-budget shed (which is planned before the run, so a reader can tell "we chose not to" from "we ran out"), and appear in `perAgent` with `failed: "budget"` rather than as agents that ran and found nothing. The sheds are read back out of the ledger when the result is assembled, so the record and the disclosure cannot disagree.

## Rollback

`REVIEW_SPEND_ENFORCEMENT=proxy-only` measures without enforcing: the record still reports `crossed`, still lists the sheds, and the notice is loud. A rolled-back run is not a blind run. The record also names which enforcement was in force, per the standing rule that where enforcement changes over time the artifact says which one produced it.

## Telemetry

`DispatchResult.spend` (ceiling, reserve, spend, overshoot, sheds, enforcement) is the run/cost/outcome record, staged in both `dispatch-result.json` and the `out/` copy the artifact upload carries. Deliberately a schema rather than an ad-hoc log line, because it is the shape a second consumer would need.

## What this does NOT do

It does not restore the *location* property of the proxy cap. awf's cap is un-weakenable because it lives in a container image no Khan repo can edit; this ceiling lives in a repo whose PRs this reviewer reviews. The provider-side workspace limit that would have replaced that property is not available, so `CODEOWNERS` on the ledger (two owners: one stalls when away, none enforces nothing) plus required review is the whole mitigation. That trade is recorded in the module itself, not just here.

## Tests

1691 pass. 13 on the ledger's arithmetic and abort wiring, 4 integration tests on the run: one dispatch paid for and the rest refused under a tight ceiling, the whole roster running under the rollback, the record present in both staged files, and the enforcement named.

### DIFF
--- a/.changeset/review-spend-ceiling.md
+++ b/.changeset/review-spend-ceiling.md
@@ -1,0 +1,61 @@
+---
+"review": minor
+---
+
+review: enforce the run's spend ceiling in code, in real dollars
+
+Dispatch now carries a budget ledger. Work may start while the run's spend is
+under a dispatch budget of `CEILING_USD - LANDING_RESERVE_USD`; crossing it
+refuses further dispatches, aborts the agents already in flight, and discloses
+every shed as a note line rather than as a silently missing dimension.
+
+Why in code rather than at the proxy. The only ceiling today is gh-aw's
+api-proxy credit cap (`maxAiCredits: 2500` in the compiled lock), which is
+denominated in list-price credits rather than the dollars Khan pays, cannot shed
+gracefully (it fails the run), and disappears with the proxy when the migration
+completes. This supersedes the intent of #314, which tried to fix the
+denomination at the proxy and was blocked on a gh-aw version; the proxy cap
+stays as a coarse list-price backstop while it exists.
+
+The numbers are derived, not chosen, and a test pins them so a change to live
+spend behaviour is a review conversation: **$12.50** ceiling, from the lock's
+2500-credit allowance minus the detection pass's measured ~$0.27 which this
+ledger does not govern, against a measured ~$0.50 for a full case's sub-agents;
+**$1.50** landing reserve, from the validator dispatch's measured ~$0.30 plus
+the reconciler. The reserve is the part that is easy to omit and expensive to
+miss: a ceiling with nothing held back turns an over-budget run into a wasted
+one, because the run can no longer validate or post what it already paid to
+find. Each consumer compiles its own lock, so a consumer with a different
+allowance wants a different ceiling.
+
+Enforcement is per turn where it can be. The runner asks the ledger at each turn
+boundary whether the agent may continue, so a single heavy reviewer cannot
+outspend a whole wave before anyone notices, and stopping happens at a turn
+boundary rather than by throwing: an agent stopped for budget has usually
+already found something worth keeping. The ledger has exactly one writer (the
+dispatcher, once per completed attempt, so retries and refusal fallbacks each
+pay), and the per-turn check is a read-only probe against it, which is what
+keeps the accounting from double counting. Concurrency is approximated the way
+the investigation cap approximates it: an in-flight agent cannot see its
+siblings' unsettled spend, so a wave can overshoot by at most the number of
+agents running at once, and `overshootUsd` measures that rather than assuming it
+away.
+
+`REVIEW_SPEND_ENFORCEMENT=proxy-only` is the rollback, and it measures without
+enforcing: the record still reports `crossed`, still lists the sheds, and the
+notice is loud, so a rolled-back run is not a blind run. The record names which
+enforcement was in force, per the standing rule that where enforcement changes
+over time the artifact says which one produced it.
+
+The record itself (`DispatchResult.spend`: ceiling, reserve, spend, overshoot,
+sheds, enforcement) is the run/cost/outcome telemetry shape, staged in both
+`dispatch-result.json` and the `out/` copy the artifact upload carries. It is
+deliberately a schema rather than an ad-hoc log line, because the same shape is
+what a second consumer would need.
+
+One property this does NOT restore, recorded in the module rather than implied:
+awf's cap is un-weakenable because it lives in a container image no Khan repo
+can edit, and this ceiling lives in a repo whose PRs this reviewer reviews. The
+provider-side workspace limit that would have replaced that property is not
+available, so CODEOWNERS on the ledger (naming two owners, since one stalls when
+away and none enforces nothing) plus required review is the whole mitigation.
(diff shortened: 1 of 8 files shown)


jwbron:
Addressed all feedback in 70fcbb1 (base sync with #305 in 1b2ead2):

- spend-ledger.ts:235 (blocking, completer mislabeled): recordSpend no longer sheds the completing agent. The agents the crossing actually cuts are disclosed through the new ledger.recordAborted, fed by the dispatcher: a budget-stopped completion always, a thrown attempt only when the ledger's signal killed it. A budget cut recorded upstream as generic 'unavailable' is upgraded to cause budget, closing the wrong-agents disclosure defect the comment named.
- dispatch.ts:347 (blocking, reserve cannot fund landing): ledger.enterLanding() at Phase 3 moves mayDispatch/wouldCross to the full ceiling and swaps in a fresh abort signal (the runner now accepts a signal thunk, resolved per request), so the validator spends the reserve instead of being refused or born aborted. Tests pin the whole sequence: crossed fan-out, live landing gate, ceiling still binding.
- dispatch-runner-pi.ts:818 and :684 (blocking, untested seams): both pinned. onTurnCost -> shouldStopAfterTurn -> stoppedForBudget (with probe values and turn count), the continue path, the aborted-later rejection, and the pre-aborted linkage.
- spend-ledger.ts:258 (rollback false sheds): proxy-only mayDispatch no longer records sheds for dispatches that ran; the run carries an explicit crossed-with-enforcement-off note instead.
- spend-ledger.ts:138 (env type): now Record<string, string | undefined>.
- CODEOWNERS:16 (question): accurate as asked: the constants live in the owned file but enforcement is wired in dispatch.ts/dispatch-calls.ts/dispatch-runner-pi.ts, which CODEOWNERS does not cover. Extending the rule to those files is a one-line change; flagging for the PR author to decide scope.
- spend-ledger.ts:57 (thought, CEILING_USD duplicates the runtime cap): deliberate while both exist; #305 deletes the proxy and its discovered cap, at which point this constant is the only ceiling. Reconciling the two before that lands would couple this PR to credit-cap.ts for one release of overlap.

jwbron:
CI note, unrelated to this PR's diff: the agent job on 1b2ead2 (run 31745225791) died from the firewall api-proxy's cache-miss guard, not from the branch. The artifact log shows `403 {"type":"max_cache_misses_exceeded","message":"Maximum consecutive cache misses exceeded (8 / 5)"}` after a healthy run (all 11 out-files staged, 16 turns, $1.06). The gh-aw claude-harness misclassifies this 403 as an authentication failure ("authentication failed - not retrying"), turning a retryable proxy guard into a hard fail; #294 hit the same signature earlier today and its rerun passed. Two follow-ups worth filing upstream against gh-aw: the 403 misclassification, and whether maxCacheMisses: 5 is too tight when the API is shedding cache under load.

jwbron:
Closing alongside #305: this threads its ledger into the runner that PR built, and that runner is being deleted rather than merged (see #305's closing comment for the record). The decisions here survive as requirements for the redo invoker ([KORE-2504](https://khanacademy.atlassian.net/browse/KORE-2504)): the $12.50 ceiling, single-writer ledger, abort-in-flight on crossing. Whether this ledger code transplants or gets rewritten is that work's call; per-turn cost visibility under headless pi is exactly the open spike question ([KORE-2503](https://khanacademy.atlassian.net/browse/KORE-2503)). The real-rate metering piece already landed separately in #314.


[KORE-2504]: https://khanacademy.atlassian.net/browse/KORE-2504?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[KORE-2503]: https://khanacademy.atlassian.net/browse/KORE-2503?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

